PR #2, opened 2019-01-15
[CHEAP-443] - Button changes

fpunny:
- **Added icon attribute to button**
   - Described within the storyboard
- **Changed proptypes to be node instead of text**
   - This allows us to do things like add anchors and icons within text without issues

doggo

### DIFF
--- a/src/components/inputs/Input.js
+++ b/src/components/inputs/Input.js
@@ -59,13 +59,8 @@ export const Input = ({
     }) => {
     return (
         <div className={className}>
-<<<<<<< HEAD
             {title ? <FormTitle htmlFor={name}>{title}</FormTitle> : null}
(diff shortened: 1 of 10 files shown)


src/components/inputs/Input.js
```diff
@@ -59,8 +59,13 @@ export const Input = ({
     }) => {
     return (
         <div className={className}>
+<<<<<<< HEAD
+            {title ? <FormTitle htmlFor={name}>{title}</FormTitle> : null}
+            <FormInputField type={type} placeholder={placeholder} name={name} onChange={onChange} value={value}></FormInputField>
+=======
```
ralph-dev: Merging issue here

fpunny: My bad T.T

PR #3, opened 2019-01-21
Hack Day

fpunny:
Hday changes


src/components/inputs/Radio.js
```diff
@@ -0,0 +1,95 @@
...
+const Check = styled.input`
+    width: 100%;
+    height: 100%;
+    opacity: 0;
+    margin: 0;
+    z-index: 1;
+    position: absolute;
+
+    &:checked ~ div {
+        border-color: #ED242A;
```
junthehacker: Should be variables.

PR #11, opened 2019-01-25
Fixed checkbox related components

fpunny:
All yours :p


bholagabbar:
@ralph-dev @fpunny Can we please get rid of/rename the .storybook folder in root and remove .DS_Store while adding it to .gitignore ?

fpunny:
@bholagabbar 
- We need the .storybook folder because it has addons and config on how things are loaded
- Also removed and ignored the .DS_Store

PR #20, opened 2019-02-01
CHEAP-480 Popup onClose

ismailmohammad:


src/components/containers/Popup.js
```diff
@@ -109,12 +110,12 @@ export class Popup extends Component {
     }
 
     render() {
-        const { className, name, padding, width = '80%', height = '80%', children } = this.props;
+        const { className, name, padding, width = '80%', height = '80%', children, onClose = this.hide } = this.props;
```
fpunny: I dont think this is the right way to modify it. onClose is a handler to the close event, but this makes it optional to close (which is questionable why you want this). So consider this instead

ie.

hide = () => {
  const { onClose } = this.props;
  popup.hide(this.props.name);
  if (onClose) onClose();
}

src/components/inputs/Input.js
```diff
@@ -26,6 +26,7 @@ const FormInputField = styled.input`
     margin: 5px 0;
     opacity: 0.8;
     transition: opacity ease-in-out 300ms;
+    font-size: 14px;
```
fpunny: try to use rem if possible :p (this is fine too)

PR #35, opened 2019-02-22
CEM-18 Create this component in React, using styled components.

JingyiCao:
Created animated card object

### DIFF
--- a/src/components/cards/cards.js
+++ b/src/components/cards/cards.js
@@ -1,0 +1,99 @@
+import React                                          from 'react';
+import styled, {css}                                  from 'styled-components';
+import PropTypes                                      from 'prop-types';
+import {PRIMARY_COLOUR, PRIMARY_FONT} from "../variables";
+import CircularProgressbar from 'react-circular-progressbar';
+import 'react-circular-progressbar/dist/styles.css';
+import 'w3-css/w3.css';
+
+
+const Card = styled.div`
+box-shadow: 0 4px 8px 0 rgba(0,0,0,0.2);
+border-radius: 5px;
+height:500px;
+width:500px;
+margin: 40px;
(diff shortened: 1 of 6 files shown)


src/components/cards/cards.js
```diff
@@ -0,0 +1,148 @@
...
+
+const HeadingOneStyled = styled.h1`
+    font-family: ${PRIMARY_FONT};
+    width: 300px;
+    margin-top: 30px;
+    margin-left:30px;
+    display: inline-block;
+    font-weight: ${props => props.bold && "bold"};
+`;
+const HeadingThreeStyled = styled.h3`
```
ralph-dev: You should be using the existing HeadingOne in this repo loool

src/components/cards/cards.js
```diff
@@ -0,0 +1,148 @@
...
+                            percentage={percentage}
+                            text={`${percentage} mins`}
+                            styles={style}
+                        />
+                    </CircleContainer>
+                    <HeadingThreeStyled bold={bold}>
+                        {textTwo? textTwo: children}
+                    </HeadingThreeStyled>
+                </SliderDiv>
+                <SliderDiv>
```
ralph-dev: Remove this whole thing looks bad, should just be two texts and one circle

src/components/cards/cards.js
```diff
@@ -0,0 +1,148 @@
...
+    textThree,
+    textFour,
+    children,
+    bold
+}) => {
+    return (
+        
+            <Card>
+                <SliderDiv>
+                    <HeadingOneStyled bold={bold}>
```
ralph-dev: What you should do is replace this whole thing with {children} that way you can just pass the internal elements of the card and style them anyway you wish

PR #37, opened 2019-02-24
CHEAP-553 Add darkmode for Date Picker

kevin51jiang:
Added darkmode for date picker, and started the foundations needed for an app-wide darkmode.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2681,7 +2681,7 @@
     },
     "babel-plugin-syntax-jsx": {
       "version": "6.18.0",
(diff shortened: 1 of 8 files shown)


fpunny:
Hey Kevin,

Thanks for your PR. Your idea very good, as it can help us open a door to better and easier customization in an application level in real-time! However, due to how specific it is, it would not be directly implemented.

Instead however, it would be introduced in our Layout PR, where it would sit in a global context / style provider. (/components/preview/Global && /components/preview/Theme)

Once again thanks for your idea, and if you have any feedback. Feel free to let us know!

Frederic

PR #38, opened 2019-02-25
CHEAP-552 Created Layouts

fpunny:
# Changes

### Layouts!
Used as an internal component for sharing styles between components **(DO NOT EXPORT)**
They are stored in the "_helpers" folder, with an underscore to remind users that its internal
- Created TextLayout to unify overall look of all text such as SmallText, Paragraph, and Heading
- Created InputLayout to unify frame of inputs (such as error, title, and description)

### Migrations
- Heading is now one SC component for better maintainability (Pls deprecate old way, they are still here so we don't implode)
- All text uses new TextLayout
- Migrated Input with slightly fresher look and feel

### Issues
- Extended PropTypes don't show
- SC inheritance killed my ref, so error animation isn't perfect (Anything more than 100px would fail)

For Ralph: Lack of pictures due to no real visual change


bholagabbar:
This is a big PR and I don't know the change exactly lol. Waiting on @junthehacker to merge it. Could do with an image of what's changed @fpunny

fpunny:
There isn't really any visual changes, but i agree, don't merge it for now. We might need to create a "legacy" version of old inputs (text (other than heading) wouldnt change, i literally only just pumped it with way more customizations) because now we would be using a consistent naming scheme for certain key attributes

@bholagabbar 
@ralph-dev

fpunny:
image
image
image
image

with the power of layouts, input literally looks and behaves the exact same as select

fpunny:
**Updates**

- Added global styles with themes with <Global/>
  - Have to transfer all variables to theme variables (migration for this should be literally 1 second, plus very very useful)
- Standardized date picker
- Added new container layout, which allows for keyboard navigation and standardized labels, descriptions, and spacing of items

PR #40, opened 2019-03-01
CEM-18 Create this component in React, using styled component

JingyiCao:
Created Card with animated circle with knobs 
image


bholagabbar:
@JingyiCao why do you keep closing pull requests and opening new ones ? You can make corrections in the same one.

PR #45, opened 2019-07-20
[HTV4-29] Create Card component

leungbyr:
image


fpunny:
This looks really good actually (Error is mostly from typescript stuff)

PR #51, opened 2020-01-29
Migrated Image from v1 to v2

nikitasheremet:
Moved the Image Component from V1 to V2 and converted file to tsx. Added to StoryBook.

Image component is an Upload Image button.


src/Inputs/Button.tsx
```diff
@@ -17,17 +17,19 @@ export interface ButtonProps extends MainInterface, ResponsiveInterface {
     loading?: boolean;
     full?: boolean;
     onClick?: React.MouseEventHandler;
+    disabled?: boolean;
```
fpunny: oof my bad

src/Inputs/Image.tsx
```diff
@@ -0,0 +1,160 @@
...
+    extends MainInterface,
+        ResponsiveInterface,
+        ImplicitPropsInterface {
+    maxWidth?: number | undefined;
+    accept?: string;
+    aspect?: number;
+    onChange?: Function;
+}
+
+const Drop = styled.input`
```
fpunny: I would move these styled.xxx to the bottom

src/Inputs/Image.tsx
```diff
@@ -0,0 +1,160 @@
...
+            setImage(URL.createObjectURL(file));
+            modal[1](true);
+        }
+    }, []);
+
+    const onClose = useCallback(() => {
+        URL.revokeObjectURL(image);
+        setCrop({ aspect });
+        img.current = { x: '', y: '', width: '', height: '' };
+        setImage(null);
```
fpunny: change this to undefined. Null is evil >:c

src/Inputs/Image.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+const Drop = styled.input`
+    ${position('absolute', 'auto', 0)}
+    opacity: 0;
+    cursor: pointer;
+    height: 100%;
+    width: 100%;
+`;
+
+const Container = styled.div<ImageProps>`
```
fpunny: These styles are very outdated, perhaps we should make this a button

src/Themes/ThemeTemplate.ts
```diff
@@ -146,6 +147,7 @@ const Theme: ThemeTemplateInterface = {
         page: 600,
     },
     depth: ['none', '0 1px 3px rgba(0,0,0,0.2)', '0 2px 6px rgba(0,0,0,0.3)'],
+    shadows: ['0 1px 3px rgba(0,0,0,0.2)', '0 3px 8px rgba(0,0,0,0.4)'],
```
fpunny: This is the same as depth

src/Inputs/Image.tsx
nikitasheremet: Make the Container component a button instead of a div?

PR #53, opened 2020-02-04
Upload Button - fix bug

nikitasheremet:
Fixed bug where the wrong part of the cropped image was being returned in the upload button.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,94 +1,94 @@
 {
-  "name": "@cheapreats/react-ui",
-  "version": "2.3.17",
-  "description": "React UI library for CheaprEats",
-  "main": "dist/index.js",
-  "files": [
-    "dist"
-  ],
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,94 +1,94 @@
...
-    "dist",
-    ".storybook",
-    ".deploy"
-  ],
-  "dependencies": {
-    "@types/react-image-crop": "^8.1.2",
-    "react-image-crop": "^8.5.0"
-  }
+    "name": "@cheapreats/react-ui",
+    "version": "2.3.20",
```
ralph-dev: Why did you use 20 instead of 18?

nikitasheremet: On npm it said .19
image

ralph-dev: I figured ok it’s good then to merge!

PR #54, opened 2020-02-11
Excel Options Component

nikitasheremet:
Build Excel Options component as per design: https://wolfbeacon.atlassian.net/wiki/spaces/CW/pages/1018789892/Excel+Download+Options+Component
Had to edit some logic in Date Picker component to allow for Datepicker to render without a default value.

### DIFF
--- a/src/Containers/Tag.tsx
+++ b/src/Containers/Tag.tsx
@@ -6,11 +6,13 @@ import { Main } from '@Utils/BaseStyles';
 
 export interface TagProps {
     icon?: React.ForwardRefExoticComponent<React.RefAttributes<SVGSVGElement>>;
+    onClick?: Function;
 }
 
 export const Tag: React.FC<TagProps> = ({
     icon = Times,
     children,
+    onClick = () => {},
     ...props
 }): React.ReactElement => {
     return (
--- a/src/Containers/Tag.tsx
+++ b/src/Containers/Tag.tsx
@@ -6,17 +6,17 @@ import { Main } from '@Utils/BaseStyles';
 
 export interface TagProps {
     icon?: React.ForwardRefExoticComponent<React.RefAttributes<SVGSVGElement>>;
-    onClick?: Function;
+    onClick?: () => void;
 }
 
 export const Tag: React.FC<TagProps> = ({
     icon = Times,
     children,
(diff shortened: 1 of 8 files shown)


src/Inputs/ExcelOptions.tsx
```diff
@@ -0,0 +1,243 @@
...
+                >
+                    {keyToHeader(headersInSelect).map((header, index) => (
+                        <option key={header} value={header}>
+                            {header}
+                        </option>
+                    ))}
+                </Select>
+            </div>
+            <div>
+                <p>Date Picker</p>
```
ralph-dev: CONSTANTS for From and To

src/Containers/Tag.tsx
```diff
@@ -6,15 +6,17 @@ import { Main } from '@Utils/BaseStyles';
 
 export interface TagProps {
     icon?: React.ForwardRefExoticComponent<React.RefAttributes<SVGSVGElement>>;
+    onClick?: () => void;
```
fpunny: I know this is the way I do it, but look into extending TagProps with HTMLDivElement props, that would be better

src/Inputs/Datepicker/Datebox.tsx
```diff
@@ -13,8 +13,11 @@ import { WEEKDAYS, MONTHS } from '@Utils/Constants';
 import { Button } from '../Button';
 
 const SIZE = 40;
-const displayDate = (date: Date): string =>
-    `${MONTHS[date.getMonth()]}, ${date.getFullYear()}`;
+
+const displayDate = (date: Date | undefined) =>
+    `${MONTHS[date ? date.getMonth() : new Date().getMonth()]}, ${
```
fpunny: Why not just make date have a default value of new Date?

src/Inputs/Datepicker/Datebox.tsx
```diff
@@ -64,14 +67,16 @@ const buildCalendar = (
 
 export interface DateboxProps {
     changePage: (change?: number) => React.MouseEventHandler;
-    selectedDate: Date;
+    selectedDate: Date | undefined;
```
fpunny: ```selectedDate?: Date;```

PR #59, opened 2020-02-21
Image Upload Button

nikitasheremet:
Fixed issue where the Image component was returning the URL of the blob rather than the base64 data


stories/Inputs/Image.stories.js
```diff
@@ -13,6 +13,7 @@ storiesOf('Image', module)
         return (
             <div>
                 <Image onImageReturn={data => addImage(data)} />
+                {console.log(src)}
```
ralph-dev: remove this.

PR #60, opened 2020-02-27
Emoji Picker

nikitasheremet:
A component that will allow a vendor to pick an emoji to use with their loyalty program.


src/Inputs/EmojiPicker.tsx
```diff
@@ -0,0 +1,53 @@
...
+    extends MainInterface,
+        ResponsiveInterface,
+        ImplicitPropsInterface {
+    onChange: Function;
+    showEmoji: boolean;
+    text: string;
+}
+
+export const EmojiPicker: React.FC<EmojiPickerProps> = ({
+    onChange = () => {},
```
ralph-dev: Isn’t this => void? Can we double check this

PR #63, opened 2020-03-16
CEENG-253 Allow footer to passed into Navigation

nikitasheremet:
Updated Navigation component to allow a footer prop to be passed into it. 

image

### DIFF
--- a/src/Containers/Navigation.tsx
+++ b/src/Containers/Navigation.tsx
@@ -35,6 +35,7 @@ export interface NavigationProps extends MainInterface, ResponsiveInterface {
     tag?: React.ReactElement | string;
     theme: DefaultTheme;
     label?: string;
+    footer?: string | number | React.FC;
 }
 const _Navigation: React.FC<NavigationProps> = ({
     params = {},
@@ -43,6 +44,7 @@ const _Navigation: React.FC<NavigationProps> = ({
     children,
     theme,
     label,
+    footer,
     ...props
 }): React.ReactElement => (
     <Container {...props}>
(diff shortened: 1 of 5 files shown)


fpunny:
Might as well do header while your at it xD but lgtm

src/Containers/Navigation.tsx
```diff
@@ -35,6 +35,7 @@ export interface NavigationProps extends MainInterface, ResponsiveInterface {
     tag?: React.ReactElement | string;
     theme: DefaultTheme;
     label?: string;
+    footer?: string | number | React.FC;
```
ralph-dev: Isn't this only React Element based on the code?

nikitasheremet: Based on its current use case yes, but I just made it as flexible as possible so that anything could be passed as a footer if necessary. I can change it to just React.FC if you think that makes more sense?

ralph-dev: Yeah but does it actually support Text being passed in?

ralph-dev: Like If i pass in a number does the file have <Text>{children}</Text>
because if you support React.FC that will collide with the above ^

fpunny: hmm good catch, should be ```React.ReactElement```

ralph-dev:
LGTM

src/Containers/Navigation.tsx
ralph-dev: @nikitasheremet

nikitasheremet: So leave string, and number and replace React.FC with ReactElement? Or have just ReactElement? @ralph-dev

ralph-dev: Just have react.reactelement

PR #64, opened 2020-03-16
CEENG-255 Navigation Footer

nikitasheremet:
Created a Navigation Footer component as it is reused by Dashboard and Terminal

image

### DIFF
--- a/src/Containers/NavigationFootnote.tsx
+++ b/src/Containers/NavigationFootnote.tsx
@@ -10,8 +10,8 @@ export interface NavigationFootnoteProps
     extends MainInterface,
         ResponsiveInterface,
         ImplicitPropsInterface {
-    url?: string;
-    text?: string;
+    url: string;
+    text: string;
 }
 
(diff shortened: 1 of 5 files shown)


ralph-dev:
Close this PR it should be V2 branch not V1?

src/Containers/NavigationFootnote.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { Mixins } from '@Utils';
+import { Compass } from 'styled-icons/fa-solid/Compass';
+import { Paragraph as P } from '../Text';
+
+export interface NavigationFootnoteProps
+    extends MainInterface,
+        ResponsiveInterface,
+        ImplicitPropsInterface {
+    url?: string;
+    text?: string;
```
ralph-dev: This should be required fyi

ralph-dev: Because if blank will look weird.

PR #65, opened 2020-03-18
CEENG 222 UI

nikitasheremet:
Added Image Carousel Component

image

### DIFF
--- a/src/Containers/ImageCarousel.tsx
+++ b/src/Containers/ImageCarousel.tsx
@@ -1,0 +1,88 @@
+import React from 'react';
+import { Times } from 'styled-icons/fa-solid/Times';
+import { flex, position, transition, scroll } from '@Utils/Mixins';
+import styled from 'styled-components';
+import { MainInterface, ResponsiveInterface } from '@Utils/BaseStyles';
+import { ImplicitPropsInterface } from '@Utils/Hooks';
+
+export interface ImageCarouselProps
+    extends MainInterface,
+        ResponsiveInterface,
+        ImplicitPropsInterface {
+    imageData: string[];
+    onClick: Function;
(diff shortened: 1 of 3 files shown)


src/Containers/ImageCarousel.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { MainInterface, ResponsiveInterface } from '@Utils/BaseStyles';
+import { ImplicitPropsInterface } from '@Utils/Hooks';
+
+export interface ImageCarouselProps
+    extends MainInterface,
+        ResponsiveInterface,
+        ImplicitPropsInterface {
+    imageData: string[];
+    onClick: Function;
+    altText: string;
```
ralph-dev: Doesn’t this need to be an array?

src/Containers/ImageCarousel.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <Items {...props}>
+            {imageData.map(image => (
+                <Item key={image} onClick={() => onClick(image)}>
+                    <Overlay>
+                        <Icon /> Delete
+                    </Overlay>
+                    <img
+                        width={width}
+                        height={height}
+                        alt={altText}
```
ralph-dev: This will have the same alt for every image is that right?

src/Containers/ImageCarousel.tsx
nikitasheremet: It is? It is an array of strings

src/Containers/ImageCarousel.tsx
nikitasheremet: Yeah I was thinking it would just say either Store Image, or Store Logo

PR #67, opened 2020-03-18
CEENG-262

nikitasheremet:
Added prop to pass in component above the list items


src/Containers/List.tsx
```diff
@@ -22,6 +22,7 @@ export interface ListProps extends MainInterface, ResponsiveInterface {
     header: Function;
     footer: Function;
     label: string;
+    alternateContent?: React.ReactElement;
```
ralph-dev: Rename to StickyTopContent

PR #68, opened 2020-03-19
CEENG-263 Add MaskedInput component

junthehacker:
Masked input takes in a mask function, and uses abstracted value/onChange handlers to handle events.

```
<MaskedInput
         name="demo"
         label={text('Enter a value', 'Enter a value')}
         placeholder={text('Enter a value', 'Enter a value')}
         description={text('Enter a value', 'Enter a value')}
         realValue={value}
         onRealValueChange={setValue}
         mask={s => '$' + s}
/>      
```

There are two preset masks

```
DOLLAR_FORMAT_MASK
PERCENT_FORMAT_MASK
```

there are also no error handling because input already have props to pass-in errors, applications should have its own logic to handle errors (for example you might actually want to enable 120%, I don't want to restrict that, also you might still want to have advanced custom error handling, that cannot be baked into SDK).

**Want approval from all**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 .deploy
 .awcache
 dist
+.idea
--- a/src/Fragments/InputFragment.tsx
+++ b/src/Fragments/InputFragment.tsx
@@ -6,14 +6,15 @@ import { ResponsiveInterface, MainInterface } from '@Utils/BaseStyles';
 interface InputFragmentProps extends ResponsiveInterface, MainInterface {
     disabled?: boolean;
     placeholder?: string;
-    onChange?: React.ChangeEventHandler<HTMLElement>;
+    onChange?: React.ChangeEventHandler<HTMLInputElement>;
     value?: string | number;
     error?: boolean;
     success?: boolean;
     children?: React.ReactNode;
-    onFocus?: React.FocusEventHandler<HTMLElement>;
-    onKeyDown?: React.KeyboardEventHandler<HTMLElement>;
-    onClick?: React.MouseEventHandler<HTMLElement>;
+    onFocus?: React.FocusEventHandler<HTMLInputElement>;
+    onBlur?: React.FocusEventHandler<HTMLInputElement>;
+    onKeyDown?: React.KeyboardEventHandler<HTMLInputElement>;
+    onClick?: React.MouseEventHandler<HTMLInputElement>;
     className?: string;
 }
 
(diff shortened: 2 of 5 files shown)


ralph-dev:
Have Presets for Dollar and Percentage becuase you need UI to react correctly

junthehacker:
@ralph-dev 
@fpunny 
I need guidance on how to actually export the preset, it does not make sense to bake preset into component, so I exported two const functions, not sure if that is the best idea.

ralph-dev:
In regards to what purpose? The use of tree shaking or something else?

junthehacker:
> In regards to what purpose? The use of tree shaking or something else?

Just look at the code, does it make sense?

ralph-dev:
Why Not use a typescript enum in your component to map accordingly so user importing component will know what masks are available

junthehacker:
> Why Not use a typescript enum in your component to map accordingly so user importing component will know what masks are available

Because the whole point is to allow you to have custom masks, so how do you support both presets and custom mask functions?

fpunny:
@junthehacker 

We should be able to just export a const like MaskedInputPresets, and tree shaking would happen because the code isn't bundled

junthehacker:
> bundled

ok, so does my current export make sense? I put the presets directly inside `MaskedInput.tsx` file.

fpunny:
> > bundled
> 
> ok, so does my current export make sense? I put the presets directly inside `MaskedInput.tsx` file.

Yes that is fine, make sure u add the export from the index.ts in the Input folder

ralph-dev:
> > Why Not use a typescript enum in your component to map accordingly so user importing component will know what masks are available
> 
> Because the whole point is to allow you to have custom masks, so how do you support both presets and custom mask functions?

Easily. Use 
maskFunction: PresetMaskTypes | () => string

fpunny:
> > > Why Not use a typescript enum in your component to map accordingly so user importing component will know what masks are available
> > 
> > 
> > Because the whole point is to allow you to have custom masks, so how do you support both presets and custom mask functions?
> 
> Easily. Use
> maskFunction: PresetMaskTypes | () => string

Thats not a enum...

junthehacker:
> > > Why Not use a typescript enum in your component to map accordingly so user importing component will know what masks are available
> > 
> > 
> > Because the whole point is to allow you to have custom masks, so how do you support both presets and custom mask functions?
> 
> Easily. Use
> maskFunction: PresetMaskTypes | () => string

Doesn't make sense, you cannot have enum with functions..

junthehacker:
@ralph-dev I just find it stupid to pass in an enum, since that will require us to bundle the mask function inside the component, adding unnecessary size.

Instead, it makes more sense to export the function, since if you don't need the preset, it will not be included into the prod build.

junthehacker:
@ralph-dev @fpunny refactored and pushed new code

PR #70, opened 2020-03-27
Terminal v2 ts fixes

rajkumarshahu:
@ralph-dev 
I have added type to the props for react UI for Heading component. Please review it and give me a feedback.


src/Text/Heading.tsx
```diff
@@ -3,10 +3,19 @@ import { TextLayout, TextLayoutProps } from '@Layouts/TextLayout';
 
 export interface HeadingProps extends TextLayoutProps {}
 
+const elements = {
```
ralph-dev: Use Enums!

PR #72, opened 2020-03-27
Terminal v2 ts fixes

rajkumarshahu:
@ralph-dev 
@nikitasheremet 
Add type to the props for react UI for Heading component.

### DIFF
--- a/src/Text/Heading.tsx
+++ b/src/Text/Heading.tsx
@@ -3,14 +3,14 @@ import { TextLayout, TextLayoutProps } from '@Layouts/TextLayout';
 
 export interface HeadingProps extends TextLayoutProps {}
 
-const elements = {
-    h1: 'h1',
-    h2: 'h2',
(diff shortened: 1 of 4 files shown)


src/Text/Heading.tsx
```diff
@@ -3,10 +3,19 @@ import { TextLayout, TextLayoutProps } from '@Layouts/TextLayout';
 
 export interface HeadingProps extends TextLayoutProps {}
 
+enum elements {
```
ralph-dev: I believe you want the string version so 
h1 = 'h1

src/Text/SmallText.tsx
```diff
@@ -7,9 +7,10 @@ export const SmallText: React.FC<SmallTextProps> = ({
     children,
     type = 'span',
     size = 'small',
+    color = 'text',
```
ralph-dev: default value is text??

PR #74, opened 2020-04-01
CEENG-289 Common Settings Component

nikitasheremet:
Created a common settings component between Dashboard and Terminal, by default has 2 cards Account and Info, and can accept 2 more cards.

Desktop:
image
Tablet:
image


src/Inputs/Timepicker/index.tsx
```diff
@@ -13,7 +13,7 @@ export interface TimepickerProps extends LabelLayoutProps {
     onChange?: Function;
     disabled?: boolean;
 }
-
+// @ts-ignore not sure why this is being highlighted as an issue
```
ralph-dev, line 16: Was line 25 being highlighted? I wonder if that is the issue

nikitasheremet: Let me look at the error

src/Containers/Settings.tsx
```diff
@@ -0,0 +1,205 @@
...
+    width: 100%;
+`;
+
+interface SettingsProps
+    extends MainInterface,
+        ResponsiveInterface,
+        InfoProps,
+        AccountProps {
+    loading: boolean;
+    cardOne?: React.ReactElement;
```
fpunny, line 32: Give it better names like info, account, etc.

src/Containers/Settings.tsx
```diff
@@ -0,0 +1,205 @@
...
+interface AccountProps extends MainInterface, ResponsiveInterface {
+    employee: { username: string; role: string; _id: string };
+    logout: Function;
+}
+interface DisplayItemProps extends MainInterface, ResponsiveInterface {
+    label: string;
+    value: string;
+}
+
+const Info: React.FC<InfoProps> = ({ faqName, faqLink, version }) => {
```
fpunny, line 50: Something i failed to do, which we need for every component to fix most of our TS issues is to inlude the props from the actual html element which the component is based off
https://stackoverflow.com/questions/58924179/typscript-react-extending-default-functional-component-props-gives-error

src/Containers/Settings.tsx
```diff
@@ -0,0 +1,205 @@
...
+    const state = useState(false);
+    return (
+        <SettingsCard heading="Additional Information" icon={InfoCircle}>
+            <Buttons>
+                <Button
+                    margin="5px"
+                    icon={Book}
+                    onClick={() =>
+                        window.open(
```
fpunny, line 59: If we have time, we should make an LinkButton instead

src/Containers/Settings.tsx
```diff
@@ -0,0 +1,205 @@
...
+                <Paragraph margin="15px 0" bold>
+                    {'If you require immediate assistance, please call/text: '}
+                    <a href="tel:13063411035">+1 (306) 341-1035</a>
+                </Paragraph>
+                <Button onClick={() => state[1](false)}>Close</Button>
+            </Modal>
+        </SettingsCard>
+    );
+};
+const Account: React.FC<AccountProps> = ({ employee, logout }) => {
```
fpunny, line 104: This is unholy. pls break this into multiple files

src/Containers/Settings.tsx
```diff
@@ -0,0 +1,205 @@
...
+        <>
+            <Loading loading={loading} inlineStyle={loadingStyle} {...props}>
+                <div>
+                    <Heading type="h1" bold lineHeight="1.2" margin="0 0 10px">
+                        Terminal Settings
+                    </Heading>
+                    {!loading && (
+                        <Cards>
+                            <Account employee={employee} logout={logout} />
+                            {cardTwo && cardTwo}
```
fpunny, line 167: I am confusion

src/Containers/Settings.tsx
```diff
@@ -0,0 +1,205 @@
...
+    cardThree,
+    faqName = 'FAQ',
+    faqLink,
+    version,
+    employee = { username: '', role: '', _id: '' },
+    logout,
+    ...props
+}): React.ReactElement => {
+    return (
+        <>
```
fpunny, line 158: Fragment is useless

src/Containers/Copyright.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { SmallText } from '../Text/SmallText';
+
+interface CopyrightProps extends MainInterface, ResponsiveInterface {
+    margin: string;
+    version: string | number;
+}
+
+export const Copyright: React.FC<CopyrightProps> = ({
+    margin,
+    version,
```
fpunny, line 12: Better yet, make version a theme field

fpunny, line 12: default this to npm package version

nikitasheremet: Wouldn't this just default to the react-ui package version then?

src/Containers/Settings.tsx
nikitasheremet: So we need to include the html props for every type of element that is present in the component, or the props for the outermost element (in the code above, the outermost element is SettingsCard which is a div)?

src/Inputs/Timepicker/index.tsx
```diff
@@ -13,7 +13,7 @@ export interface TimepickerProps extends LabelLayoutProps {
     onChange?: Function;
     disabled?: boolean;
 }
-// @ts-ignore not sure why this is being highlighted as an issue
+// @ts-ignore TimePicker throwing error because of withTheme
```
ralph-dev, line 16: https://github.com/mui-org/material-ui/issues/8070 does this solve it?

src/Containers/Settings.tsx
fpunny: outmost

src/Containers/Copyright.tsx
fpunny: hmm ur right, nvm then :/

src/Inputs/Timepicker/index.tsx
nikitasheremet: @fpunny Could you explain what withTheme is being used for in Timepicker?

src/Containers/Loading.tsx
```diff
@@ -9,7 +9,10 @@ import {
 } from '@Utils/BaseStyles';
 import { useTransition } from '@Utils/Hooks';
 
-export interface LoadingProps extends ResponsiveInterface, MainInterface {
+export interface LoadingProps
+    extends React.HTMLAttributes<HTMLDivElement>,
```
fpunny, line 13: LGTM owo

PR #76, opened 2020-04-08
CEENG-312 Update React UI with HTML Attributes

nikitasheremet:
I have updated as many of the components as I could with the HTML attributes. There were some issues where a component uses a "custom" onChange function but because it is also called onChange it throws an error when you try to extend an HTML interface.

### DIFF
--- a/src/Containers/ImageCarousel.tsx
+++ b/src/Containers/ImageCarousel.tsx
@@ -8,8 +8,7 @@ import { ImplicitPropsInterface } from '@Utils/Hooks';
 export interface ImageCarouselProps
     extends MainInterface,
         ResponsiveInterface,
-        // doesn't work for some reason - havent been able to find a solution
-        // React.HTMLAttributes<HTMLUListElement>,
+        Omit<React.HTMLAttributes<HTMLUListElement>, 'onClick'>,
         ImplicitPropsInterface {
     imageData: string[];
     onClick: Function;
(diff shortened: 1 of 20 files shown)


src/Containers/NavigationFootnote.tsx
```diff
@@ -9,7 +9,8 @@ import { Paragraph as P } from '../Text';
 export interface NavigationFootnoteProps
     extends MainInterface,
         ResponsiveInterface,
-        ImplicitPropsInterface {
+        ImplicitPropsInterface,
+        React.HTMLAttributes<HTMLDivElement> {
```
fpunny, line 13: Ya div would do. nav is just a specialized div

src/Containers/ImageCarousel.tsx
```diff
@@ -8,6 +8,8 @@ import { ImplicitPropsInterface } from '@Utils/Hooks';
 export interface ImageCarouselProps
     extends MainInterface,
         ResponsiveInterface,
+        // doesn't work for some reason - havent been able to find a solution
```
fpunny, line 11: Might be the onClick prop conflicting with it, remove it and see what happens?

nikitasheremet: @fpunny But dont you need the custom onClick prop because it returns the image? If I remove the onClick you won't be able to click on the image to delete it

fpunny: https://stackoverflow.com/questions/41285211/overriding-interface-property-type-defined-in-typescript-d-ts-file

KK then try one of these solutions which omit's the property so that it can be made custom

nikitasheremet: Omit works! @fpunny

PR #81, opened 2020-04-15
added ComboBox component to the library

sahandfarahani:
ComboBoxgif


src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,297 @@
...
+    const handleChange=(event:any)=>{
+        myWindow.inputValue=event.target.value;
+        setExpanded(true);
+    }
+
+    return (
+        <LabelLayout {...props}>
+            <Container>
+
+                    <InputFragment value={myWindow.inputValue}
```
ralph-dev, line 146: The indentation looks off?

src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,297 @@
...
+                                } 
+                    onClick={e=>
+                        handleChange(e)
+                        } 
+                        
+                     />
+
+                {mount && (
+                    <SelectList
+                        limit={Math.min(3, limit)}
```
ralph-dev, line 161: 3 is a magic number, use a constant NUM_VISIBLE_SELECTIONS = 3;

src/Inputs/ComboBox.tsx
sahandfarahani: sorry
fixed that

src/Inputs/ComboBox.tsx
sahandfarahani: done

src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,294 @@
...
+const _Select: React.FC<SelectProps> = ({
+    disabled,
+    value,
+    children,
+    limit = 4,
+    placeholder = '',
+    onChange = (): void => {},
+    theme,
+    name,
+    NUM_VISIBLE_SELECTIONS,
```
ralph-dev, line 45: This should be a variable not a constant?

Meaning
, numVisible = NUM_VISIBLE_SELECTIONS

and at line 21, const NUM_VISIBLE_SELECTIONS = 3;

sahandfarahani: I think I did it as mentioned now

src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,295 @@
...
+    position,
+    transition,
+    clickable,
+    styledCondition,
+    darken,
+} from '@Utils/Mixins';
+import { useTransition } from '@Utils/Hooks';
+import { LabelLayout, LabelLayoutProps,InputFragment } from '@Layouts';
+ 
+const NUM_VISIBLE_SELECTIONS=3;
```
ralph-dev, line 22:
```suggestion
const NUM_VISIBLE_SELECTIONS = 3;

```

src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,295 @@
...
+const _Select: React.FC<SelectProps> = ({
+    disabled,
+    value,
+    children,
+    limit = 4,
+    placeholder = '',
+    onChange = (): void => {},
+    theme,
+    name,
+    numVisible=NUM_VISIBLE_SELECTIONS,
```
ralph-dev, line 47:
```suggestion
    numVisible = NUM_VISIBLE_SELECTIONS,

```

src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,295 @@
...
+    ...props
+}): React.ReactElement => {
+    const [expanded, setExpanded] = useState(false);
+    const options = Children.toArray(children);
```
ralph-dev, line 51: This will run every render, should this be in a useEffect?

src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,295 @@
...
+            });
+            myWindow.inputValue=currentTarget.firstChild.nodeValue;
+        },
+        [name],
+    );
+
+    useEffect((): void | (() => void | undefined) => {
+        if (expanded) {
+            const listener = (): void => {
+                setExpanded(false);
```
ralph-dev, line 122: Where is the isMounted control?

src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,295 @@
...
+            };
+            const timer = window.setTimeout((): void => {
```
ralph-dev, line 124: Also, let's name these timers more properly

sahandfarahani: all done

src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,241 @@
...
+                window.clearTimeout(timerforkeydown);
+                window.clearTimeout(timerforclick);
+                window.removeEventListener('keydown', listener);
+                window.removeEventListener('click', listener);
+            };
+        }
+        return undefined;
+    }, [expanded]);
+
+    const handleChange=(event:any)=>{
```
nikitasheremet, line 140: (event: SyntheticEvent), import Synthetic Event from react

src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,241 @@
...
+                    value: currentTarget.getAttribute('value'),
+                    name,
+                },
+            });
+            myWindow.inputValue=currentTarget.firstChild.nodeValue;
+        },
+        [name],
+    );
+
+    useEffect((): void | (() => void | undefined) => {
```
nikitasheremet, line 113: You do not need to add any extra typings for useEffect, as far as I know

src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,241 @@
...
+            }, 10);
+            return (): void => {
+                isMounted=false;
+                window.clearTimeout(timerforkeydown);
+                window.clearTimeout(timerforclick);
+                window.removeEventListener('keydown', listener);
+                window.removeEventListener('click', listener);
+            };
+        }
+        return undefined;
```
nikitasheremet, line 137: The correct return would be return = () => isMounted = false
A useEffect returns a clean up function ( () => {} )

sahandfarahani: so I should remove isMounted=false from line130 , correct?

src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,241 @@
...
+    disabled?: boolean;
+    placeholder?: string;
+    value?: string | number;
+    theme: DefaultTheme;
+    onChange?: Function;
+    limit?: number;
+    numVisible:number;
+}
+
+const _Select: React.FC<SelectProps> = ({
```
fpunny, line 41: Why does combobox copy code from select and not use select?

src/Inputs/ComboBox.tsx
```diff
@@ -0,0 +1,241 @@
...
+    transition,
+    clickable,
+    darken,
+} from '@Utils/Mixins';
+import { useTransition } from '@Utils/Hooks';
+import { LabelLayout, LabelLayoutProps,InputFragment } from '@Layouts';
+
+const NUM_VISIBLE_SELECTIONS = 3;
+
+var options:any = [];
```
fpunny, line 23: No >:c bad

src/Inputs/ComboBox.tsx
fpunny: useMemo

src/Inputs/ComboBox.tsx
fpunny: uhhhh React.ClickEvent or something like that is better if you can

src/Inputs/ComboBox.tsx
```diff
@@ -117,19 +111,25 @@ const _Select: React.FC<SelectProps> = ({
     );
 
     useEffect((): void | (() => void | undefined) => {
+        isMounted=true;
```
ralph-dev, line 115: let isMounted

src/Inputs/ComboBox.tsx
```diff
@@ -13,13 +12,16 @@ import {
...
 import { useTransition } from '@Utils/Hooks';
 import { LabelLayout, LabelLayoutProps,InputFragment } from '@Layouts';
- 
-const NUM_VISIBLE_SELECTIONS=3;
+
+const NUM_VISIBLE_SELECTIONS = 3;
+
+var options:any = [];
+
+var isMounted=false;
```
ralph-dev, line 25: declare inside the useEffect

src/Inputs/ComboBox.tsx
fpunny: @nikitasheremet mans learning owo

fpunny: @sahandfarahani ur that is correct, but ur implementation of isMounted needs work and should also be set to false on line 137

src/Inputs/ComboBox.tsx
fpunny: unfortunately u do since we have function return being required

src/Inputs/ComboBox.tsx
sahandfarahani: > @fpunny well it gets executed whether Click or Keydown happens so maybe a more general type is better

src/Inputs/ComboBox.tsx
sahandfarahani: :)))
everything from 23 to 29?

src/Inputs/ComboBox.tsx
sahandfarahani: its mostly select, I'm going to delete the unnecessary props that were useless in the combobox

src/Inputs/ComboBox.tsx
sahandfarahani: I'm using options array in a useMemo and changing window type from any to Window & typeof globalThis

sahandfarahani: I'm using options in a useMemo hook now

src/Inputs/ComboBox.tsx
fpunny: No but why are we copying the select component code and not using the select component directly

PR #86, opened 2020-05-04
CEENG-394 Fix Relative Path and Interface Issues

nikitasheremet:
Removed @ paths for now, as I was unable to figure out a way to fix the relative paths in the declaration files. Everything is working now, and this will solve the ts issues with the components in the Dashboard and Terminal


ralph-dev:
@nikitasheremet look at slack comments about using babel to keep the alias alive post compilation or Call on Tandem / check customer app

PR #87, opened 2020-05-07
added searchbar to the library

sahandfarahani:
searchbar

### DIFF
--- a/src/Inputs/SearchBar.tsx
+++ b/src/Inputs/SearchBar.tsx
@@ -122,7 +122,6 @@ export const SearchBar: React.FC<SearchBarProps> = ({
                 window.removeEventListener('click', listener);
             };
         }
-        return undefined;
     }, [expanded]);
 
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
--- a/src/Inputs/SearchBar.tsx
+++ b/src/Inputs/SearchBar.tsx
@@ -22,18 +22,20 @@ import { useTransition } from '@Utils/Hooks';
 import { Search } from 'styled-icons/fa-solid';
 import styled from 'styled-components';
 
+const MAX_VIEWING_LIMIT = 4;
+const REVERSE = '-1';
+
 export interface SearchBarProps extends LabelLayoutProps {
     suggestiveOptions?: string[];
     value?: string | number;
     onChange?: Function;
     limit?: number;
-    numVisible: number;
 }
 
 export const SearchBar: React.FC<SearchBarProps> = ({
     suggestiveOptions,
     value,
-    limit = 4,
+    limit = MAX_VIEWING_LIMIT,
     onChange = (): void => {},
     name,
     ...props
@@ -43,13 +45,6 @@ export const SearchBar: React.FC<SearchBarProps> = ({
     const [, mount, animation] = useTransition(expanded);
     const refSelectList = useRef() as RefObject<HTMLDivElement>;
     const [numVisibleSelection, setNumVisibleSelection] = useState(limit);
-
-    const listener = (): void => {
-        setExpanded(false);
-    };
-
-    window.addEventListener('keydown', listener, { once: true });
-    window.addEventListener('click', listener, { once: true });
 
     const createList = (
         children: string[],
@@ -81,7 +76,7 @@ export const SearchBar: React.FC<SearchBarProps> = ({
                 const order = child
                     .toLowerCase()
                     .startsWith(inputValue.toLowerCase())
-                    ? '-1'
+                    ? REVERSE
                     : '1';
                 return (
                     <SelectItem
@@ -118,10 +113,22 @@ export const SearchBar: React.FC<SearchBarProps> = ({
 
     useLayoutEffect((): void | (() => void) => {
         if (refSelectList.current?.children.length) {
-            setNumVisibleSelection(refSelectList.current?.children.length);
+            setNumVisibleSelection(refSelectList.current.children.length);
         } else {
             setNumVisibleSelection(0);
         }
+
+        const listener = (): void => {
+            setExpanded(false);
+        };
+
+        window.addEventListener('keydown', listener, { once: true });
+        window.addEventListener('click', listener, { once: true });
+
+        return (): void => {
+            window.removeEventListener('keydown', listener);
+            window.removeEventListener('click', listener);
+        };
     }, [expanded]);
 
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
--- a/src/Inputs/SearchBar.tsx
+++ b/src/Inputs/SearchBar.tsx
@@ -4,6 +4,8 @@ import React, {
     useMemo,
     Children,
     useLayoutEffect,
+    useRef,
+    RefObject,
 } from 'react';
 import {
     flex,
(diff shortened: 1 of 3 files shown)


src/Inputs/SearchBar.tsx
```diff
@@ -0,0 +1,239 @@
...
+        if (expanded) {
+            const listener = (): void => {
+                if (isMounted) {
+                    setExpanded(false);
+                }
+            };
+
+            const timerforkeydown = window.setTimeout((): void => {
+                window.addEventListener('keydown', listener, { once: true });
+            }, 10);
```
ralph-dev, line 122: Why is this 10? If you are waiting for the mount why not useLayoutEffect?

src/Inputs/SearchBar.tsx
```diff
@@ -108,31 +108,21 @@ export const SearchBar: React.FC<SearchBarProps> = ({
...
-            const timerforkeydown = window.setTimeout((): void => {
-                window.addEventListener('keydown', listener, { once: true });
-            }, 10);
-            const timerforclick = window.setTimeout((): void => {
-                window.addEventListener('click', listener, { once: true });
-            }, 10);
+            window.addEventListener('keydown', listener, { once: true });
+            window.addEventListener('click', listener, { once: true });
+
             return (): void => {
```
ralph-dev, line 120: you must only have ONE return statement btw

ralph-dev: This removeListner should run every time.

src/Inputs/SearchBar.tsx
```diff
@@ -108,31 +108,21 @@ export const SearchBar: React.FC<SearchBarProps> = ({
...
-                isMounted = false;
-                window.clearTimeout(timerforkeydown);
-                window.clearTimeout(timerforclick);
                 window.removeEventListener('keydown', listener);
                 window.removeEventListener('click', listener);
             };
         }
-        // eslint-disable-next-line no-return-assign
-        return (): false => (isMounted = false);
+        return undefined;
```
ralph-dev, line 125: delete this line

ralph-dev:
The return statement cannot be in an if statement

sahandfarahani:
> 
> 
> The return statement cannot be in an if statement

so If I delete the If statement there wouldn't be a problem right?... cause expanded will be set to false inside the useLayoutEffect and the only way to trigger it is to set it to truth anyway, right?

ralph-dev:
> > The return statement cannot be in an if statement
> 
> so If I delete the If statement there wouldn't be a problem right?... cause expanded will be set to false inside the useLayoutEffect and the only way to trigger it is to set it to truth anyway, right?

Yep

ralph-dev:
@sahandfarahani One nit pick thing is that the combo box should not show 3 rows if there are only 2 possible options (Should only show 2 possible rows)

src/Inputs/SearchBar.tsx
```diff
@@ -0,0 +1,236 @@
...
+    const options = useMemo((): any[] => {
+        return Children.toArray(suggestiveOptions);
+    }, [expanded]);
+
+    useLayoutEffect((): void | (() => void) => {
+        if (refSelectList.current?.children.length) {
+            setNumVisibleSelection(refSelectList.current?.children.length);
+        }
+
+        const listener = (): void => {
```
ralph-dev, line 117: Shouldn't this be deleted?

sahandfarahani: @ralph-dev  but then where should I setExpanded to false when clicks or keydown events happen?

ralph-dev: Ok you are right, I thought this was a generic listeners becuase of the name (Maybe name is ExpandedListener instead?)

src/Inputs/SearchBar.tsx
```diff
@@ -0,0 +1,236 @@
...
+                child &&
+                inputValue &&
+                child.toLowerCase().search(inputValue.toLowerCase()) > -1
+            ) {
+                const val = child;
+                const selected = String(value) === val;
+                const order = child
+                    .toLowerCase()
+                    .startsWith(inputValue.toLowerCase())
+                    ? '-1'
```
ralph-dev, line 77: Abstract into REVERSE = -1

src/Inputs/SearchBar.tsx
```diff
@@ -0,0 +1,236 @@
...
+    value?: string | number;
+    onChange?: Function;
+    limit?: number;
+    numVisible: number;
+}
+
+export const SearchBar: React.FC<SearchBarProps> = ({
+    suggestiveOptions,
+    value,
+    limit = 4,
```
ralph-dev, line 36: Abstract 4 to MAX_VIEWING_LIMIT

src/Inputs/SearchBar.tsx
```diff
@@ -0,0 +1,236 @@
...
+        },
+        [name],
+    );
+
+    const options = useMemo((): any[] => {
+        return Children.toArray(suggestiveOptions);
+    }, [expanded]);
+
+    useLayoutEffect((): void | (() => void) => {
+        if (refSelectList.current?.children.length) {
```
ralph-dev, line 113: Just wondering if the ?. is nessesary since this is in a useLayoutEffect now?

src/Inputs/SearchBar.tsx
```diff
@@ -0,0 +1,236 @@
...
+    value,
+    limit = 4,
+    onChange = (): void => {},
+    name,
+    ...props
+}): React.ReactElement => {
+    const [expanded, setExpanded] = useState(false);
+    const [inputValue, setinputValue] = useState('');
+    const [, mount, animation] = useTransition(expanded);
+    const refSelectList = useRef() as RefObject<HTMLDivElement>;
```
ralph-dev, line 44:
```suggestion
    const refSelectList = useRef<HTMLDivElement>();
```
Isn't this how the TypeScript works? I might be wrong, please check this.

src/Inputs/SearchBar.tsx
```diff
@@ -0,0 +1,236 @@
...
+    const [numVisibleSelection, setNumVisibleSelection] = useState(limit);
```
ralph-dev, line 45: This is very odd, why are you setting the state to the prop value??

sahandfarahani: numVisibleSelection is not a prop anymore, I changed it to state, I think limit would do the job and I used numVisibleSelection to get the number of children inside the Selection List and the minimum of limit or  numVisibleSelection gets effect. So that when there are for example only 2 results of the suggestions the SelectList doesn't expand more than that

ralph-dev: But why is this a state? Why not just calculate this in either a ref if its never changing or inside the component that is changing every click.

src/Inputs/SearchBar.tsx
sahandfarahani: actually it doesn't work, Typescript gives error

src/Inputs/SearchBar.tsx
sahandfarahani: Well it gets the length from the visible selectitems inside the SelectList which changes with every keydown, , so it needs the SelectList to be mounted so that it can get the current number of children inside of it. thats why I put the setNumVisibleSelection inside the useLayoutEffect

src/Inputs/SearchBar.tsx
sahandfarahani: I've tried it but I'd get the error "Object is possibly null"

src/Inputs/SearchBar.tsx
ralph-dev: Ok that makes sense. Let’s merge this then.

ralph-dev:
Also use the reviewers field not assignee fiel please

PR #90, opened 2020-05-14
CEM-347 horizontal scroll list

indancisive:
HorizontalScrollList3

@ralph-dev

### DIFF
--- a/src/Containers/HorizontalScrollList.tsx
+++ b/src/Containers/HorizontalScrollList.tsx
@@ -10,6 +10,8 @@ export interface ScrollListProps extends React.HTMLAttributes<HTMLDivElement> {
     labelPadding: string;
     hoveredStyle?: Function;
     selectedStyle?: Function;
+    menuNameVary?: boolean;
+    displayDropDown?: boolean;
 }
 
 export interface ScrollListDivProps {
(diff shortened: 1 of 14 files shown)


src/Containers/HorizontalScrollList.tsx
```diff
@@ -0,0 +1,115 @@
...
+}
+
+export interface ScrollListDivProps {
+    menuWidth?: number;
+}
+
+export interface ListItemProps {
+    label: string;
+    hoveredStyle: Function;
+    selectedStyle: Function;
```
indancisive, line 24: Unsure if styling should be passed as a function that returns a string, as a standalone string, or as some other data type. Looked into React.CSSProperties but was unable to figure anything out

indancisive:
@ralph-dev @fpunny
HorizontalScrollList4
Please take a look at the two style props. Right now they are functions that return strings. Unsure if they should be just strings or something else entirely for better maintainability (possibly React.CSSProperties ?)

src/Containers/HorizontalScrollList.tsx
```diff
@@ -0,0 +1,121 @@
...
+                <DropDownDiv menuWidth={menuWidth}>
+                    <Select
+                        placeholder={menuName}
+                        onChange={({ target }: { target: any }) => {
+                            setSelectedLabel(target.value);
+                        }}
+                        value={displaySelected ? selectedLabel : menuName}
+                    >
+                        {labelArray.map(label => (
+                            <option value={label}>{label}</option>
```
ralph-dev, line 51: Where is the key prop?

src/Containers/HorizontalScrollList.tsx
```diff
@@ -0,0 +1,121 @@
...
+                </DropDownDiv>
+            )}
+            <HorizontalList>
+                {labelArray.map((label, index) => (
+                    <HorizontalListItem
+                        key={label + index}
+                        onClick={() => setSelectedLabel(label)}
+                        label={label}
+                        hoveredStyle={
+                            hoveredStyle ? hoveredStyle : defaultHoveredStyle
```
ralph-dev, line 63: this is a CSS thing not a JS thing, put an isHovering Prop and Style it in the CSS styles component.

ralph-dev: Or just use isHover from CSS?

ralph-dev:
Approve, resolve conflicts first

PR #96, opened 2020-05-25
added slider and popup

sahandfarahani:
Slider


src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,440 @@
...
-        height: 6px;
-        cursor: pointer;
-        animate: 0.2s;
-        box-shadow: 1px 1px 1px #000000, 0px 0px 1px #0d0d0d;
-        background: #3071a9;
-        border-radius: 10px;
-        border: 0.2px solid ${(props): string => props.trackColor || '#010101'};
+}): React.ReactElement => {
+    // DOM Elements
+    const bar = useRef() as RefObject<HTMLDivElement>;
```
ralph-dev, line 65: const bar= useRef<HTMLDivElement>();
Can you do this instead?

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,440 @@
...
+    const finishThumb = useRef() as RefObject<HTMLDivElement>;
+    const marksBar = useRef() as RefObject<HTMLDivElement>;
+
+    // Thumb Positions in Px
+    const [finishThumbLeft, setFinishThumbLeft] = useState(max);
+    const [startThumbLeft, setStartThumbLeft] = useState(min);
+
+    const MaxAndMinDifference = max - min;
+
+    let draggingThumb = 'Finish';
```
ralph-dev, line 77: Why are these things being computed on every render?

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,440 @@
...
+
+    // Translate a value to Pixel
+    const translateToPixels = (theValue: number): number => {
+        const pixelTranslator =
+            (bar.current?.clientWidth as number) / MaxAndMinDifference;
+        return ((theValue as number) - min) * pixelTranslator;
```
ralph-dev, line 83: do you need as Number if the type is number??

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,440 @@
...
+    function resizing(): void {
+        if (finishThumbLeftInValue) {
+            setFinishThumbLeft(translateToPixels(finishThumbLeftInValue));
+        }
+        if (startThumbLeftInValue) {
+            setStartThumbLeft(translateToPixels(startThumbLeftInValue));
+        }
+
+        if (marksBar.current && marks) {
+            const theMarks = marksBar.current.children as any;
```
ralph-dev, line 160: children are ReactNode ?

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,440 @@
...
-        width: 16px;
-        border-radius: 50%;
-        background: ${(props): string => props.buttonColor || '#ffffff'};
-        cursor: pointer;
-        -webkit-appearance: none;
-        margin-top: -5px;
+
+    window.onresize = resizing;
+
+    function onMouseMove(theevent: MouseEvent): void {
```
ralph-dev, line 176: Does ES6 notation not work for this?

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,440 @@
...
+                    </SliderThumbStart>
+                )}
+
+                <SliderThumb
+                    ref={finishThumb}
+                    disabled={disabled}
+                    hasTwoInputs={hasTwoInputs}
+                    left={finishThumbLeft}
+                    onMouseDown={(event): any =>
+                        handleMouseDown(event as MouseEvent)
```
ralph-dev, line 281: Can't you just type event: MouseEvent?

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,440 @@
...
+            </SliderBoard>
+            {marks && (
+                <SliderBoardMarks
+                    ref={marksBar}
+                    theme={theme}
+                    disabled={disabled}
+                >
+                    {marks.map(
+                        ({ key, mark }): React.ReactElement => (
+                            <div
```
ralph-dev, line 304: Where is the key?

src/Inputs/Slider.tsx
sahandfarahani: I'll get typescript errors whenever im using the props of bar

ralph-dev: I mean
useRef< HTMLDivElement >()

src/Inputs/Slider.tsx
sahandfarahani: When I put reactnode I get the ts error that reactnode is not assignable to htmlelement... If I change child type I get error that child.styles does not exist on string...

src/Inputs/Slider.tsx
sahandfarahani: I'll get ts errors that mutablerefobject<htmldivelement/undefined> is not assignable to type reconnect<htmldivelement> and ((instance:htmldivelement/null)=>voidect.

src/Inputs/Slider.tsx
sahandfarahani: Sorry I don't get this

ralph-dev: https://reactjs.org/docs/lists-and-keys.html

src/Inputs/Slider.tsx
sahandfarahani: put I put these in Useeffect or UseLayouteffect they will get undefined in the body and it's like they don't exist

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,436 @@
...
+    const selectedBar = useRef() as RefObject<HTMLDivElement>;
+    const startThumb = useRef() as RefObject<HTMLDivElement>;
+    const finishThumb = useRef() as RefObject<HTMLDivElement>;
+    const marksBar = useRef() as RefObject<HTMLDivElement>;
+
+    // Thumb Positions in Px
+    const [finishThumbLeft, setFinishThumbLeft] = useState(max);
+    const [startThumbLeft, setStartThumbLeft] = useState(min);
+
+    const MaxAndMinDifference = max - min;
```
ralph-dev, line 75: Fix

src/Inputs/Slider.tsx
```diff
@@ -10,6 +10,8 @@ import styled, { DefaultTheme } from 'styled-components';
 import { LabelLayout, LabelLayoutProps } from '@Layouts';
 import { Popup } from '../Containers/Popup';
 
+let draggingThumb = 'Finish';
```
ralph-dev, line 13: still incorrect.

src/Inputs/Slider.tsx
```diff
@@ -59,6 +62,7 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
     popupwidth = 'auto',
     popupheight = 20,
     theme,
+    maxAndMinDifference = max - min,
```
ralph-dev, line 65: Doesnt this only change when min or max changes? if so place in useeffect and watch the two variables

src/Inputs/Slider.tsx
sahandfarahani: I can use usestate for this then

sahandfarahani: But why would it be incorrect now that it does not update with every render?

src/Inputs/Slider.tsx
sahandfarahani: The thing is it never changes because min and max never change... The only reason Im using it is to make the codes shorter and more readable ... I can delete it and replace it everywhere with max-min simply

ralph-dev: @sahandfarahani Yes do that then

src/Inputs/Slider.tsx
ralph-dev: Let's jump on a call afterwards.

src/Inputs/Slider.tsx
fpunny: if you read the error message it says instance has to be null or the htmldivelement.

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,456 @@
...
+    theme?: DefaultTheme;
 }
 
 export const Slider: React.FunctionComponent<SliderProps> = ({
+    disabled = false,
+    marks,
+    min = 0,
+    max = 100,
+    hasRail,
+    hasTwoInputs = false,
```
ralph-dev, line 51: @sahandfarahani Is this hasTwoInputs supposed to be hasTwoKnobs? What is this?

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,456 @@
...
+    valueFinish = max,
```
ralph-dev, line 52: @sahandfarahani What does this variable mean?

src/Inputs/Slider.tsx
indancisive: you have to define the ref as null as the ref must either be the element object or null, not undefined
image

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,456 @@
...
-        background: #367ebd;
-    }
-    &::-moz-range-track {
+}): React.ReactElement => {
+    // DOM Elements
+    const bar = useRef() as RefObject<HTMLDivElement>;
+    const selectedBar = useRef() as RefObject<HTMLDivElement>;
+    const leftThumb = useRef() as RefObject<HTMLDivElement>;
+    const rightThumb = useRef() as RefObject<HTMLDivElement>;
+    const marksBar = useRef() as RefObject<HTMLDivElement>;
```
indancisive, line 69: it would be easier to read the later code if you name these marksBarRef, etc. just as a convention

src/Inputs/Slider.tsx
ryanarine: I think that's supposed to be one of the initial values (valueStart being the other)

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,456 @@
...
+
+    // Final Result
+    useMemo((): void => {
+        onChange({
+            values: {
+                lowValue: leftThumbPositionInValue,
+                highValue: rightThumbPositionInValue,
+            },
+        });
+    }, [rightThumbPositionInValue, leftThumbPositionInValue]);
```
indancisive, line 128: don't use useMemo for sideeffects

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,456 @@
...
+import styled, { DefaultTheme } from 'styled-components';
+import { LabelLayout, LabelLayoutProps } from '@Layouts';
+import { Popup } from '../Containers/Popup';
+
+export interface MarkProps {
+    key: number;
+    mark: string;
+}
+
+export interface ResultValues {
```
indancisive, line 18: I think some of your variables can be more clear
at a glance I have no clue what this means or what these values are for

src/Inputs/Slider.tsx
sahandfarahani: yes these are Initial Position Values of the knobs

src/Containers/Popup.tsx
```diff
@@ -0,0 +1,72 @@
...
+    height,
+    popup,
+}): React.ReactElement => {
+    return (
+        <Container popup={popup} left={left} top={top}>
+            <Rectangle width={width} height={height}>
+                {children}
+            </Rectangle>
+            <Triangle>
+                <InnerTriangle> </InnerTriangle>
```
aslamm29, line 27: might want to change it to a self closed tag if not being enclosed with child/data

stories/Inputs/Slider.stories.js
```diff
@@ -1,18 +1,91 @@
...
+        'with popup',
+        mockElement(
+            ([state, setState]) => (
+                <Slider
+                    min={20}
+                    max={50}
+                    padding="50px"
+                    hasPopup={boolean('hasPopup', true)}
+                    onChange={target => {
+                        console.log(target);
```
aslamm29, line 55: these are stories but perhaps this log may not be needed in this version here as there is a setState

stories/Inputs/Slider.stories.js
```diff
@@ -1,18 +1,91 @@
...
+            ([state, setState]) => (
+                <Slider
+                    hasTwoKnobs={boolean('twoinputs', true)}
+                    hasRail={boolean('hasRail', true)}
+                    hasPopup={boolean('hasPopup', true)}
+                    min={100}
+                    max={300}
+                    step={50}
+                    onChange={target => {
+                        console.log(target);
```
aslamm29, line 76: this one too

stories/Containers/Popup.stories.js
```diff
@@ -0,0 +1,18 @@
...
+    'with nothing',
+    () => (
+        <Popup
+            left={0}
+            top={0}
+            width={0}
+            height={0}
+            popup={true}
+            content="Content"
+        ></Popup>
```
aslamm29, line 15: marking this tag here

stories/Inputs/Slider.stories.js
```diff
@@ -1,18 +1,91 @@
...
+        mockElement(
+            ([state, setState]) => (
+                <Slider
+                    min={1}
+                    max={100}
+                    onChange={target => {
+                        setState(target);
+                    }}
+                    values={state}
+                ></Slider>
```
aslamm29, line 20: marking these and the ones below

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,384 @@
...
+         top: -21px;
+`
+            : 'top: -7px; '}
+
+    cursor: grab;
+    ${({ theme }): string => `
+            background: ${theme.colors.primary};
+        `}
+    &:active {
+        border: solid 2px #96dbfa;
```
ryanarine, line 349: add `box-sizing: border-box` so the border doesn't change the position of the knob

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,384 @@
...
+    );
+
+    const [leftKnobPosition, setLeftKnobPosition] = useState(values.leftValue);
+
+    const [isRightKnobDragging, setIsRightKnobDragging] = useState(false);
+    const [isLeftKnobDragging, setIsLeftKnobDragging] = useState(false);
+
+    const maxAndMinDifference = useMemo((): number => {
+        return max - min;
+    }, [max, min]);
```
ryanarine, line 78: Does this need to be in a use Memo? why not just `const difference = max - min`

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,384 @@
...
+            ) {
+                setRightKnobPosition(newPosition);
+            } else if (isLeftKnobDragging && newPosition < rightKnobPosition) {
+                setLeftKnobPosition(newPosition);
+            }
+        }
+    };
+
+    const onMouseUp = (): void => {
+        document.removeEventListener('mouseup', onMouseUp);
```
ryanarine, line 131: should probably add setIsRightKnobDragging(false) and setIsLeftKnobDragging(false) here

src/Inputs/Slider.tsx
ralph-dev: @ryanarine You don't want to re-calculate the difference everytime one of the other props changes though.

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,384 @@
...
+                setIsRightKnobDragging(false);
+                setIsLeftKnobDragging(true);
+                // in Case user clicks on the Bar
+                if (
+                    e.target === barRef.current ||
+                    e.target === selectedBarRef.current
+                ) {
+                    setLeftKnobPosition(roundToSteps(clickedOn));
+                }
+            }
```
ryanarine, line 169: You can make this code DRY by doing some thing like
 `const isOnRightSide = clickedOn >= ... || !hasTwoKnobs`
`setIsRightKnobDragging(isOnRightSide)`;
`setIsLeftKnobDragging(!isOnRightSide);`
`if (e.target ...) {isOnRightSide ? setRightKnobPosition(...) : setLeftKnob...}`

src/Inputs/Slider.tsx
ryanarine: @ralph-dev The overhead of useMemo probably outweighs the time it takes to subtract two numbers

ralph-dev: I agree but in that case just have the function without useMemo at all? You never want things hanging in the component because the calculation will execute almost ever re-render essentially because it's a funtional component.

src/Inputs/Slider.tsx
sahandfarahani: @ryanarine its excellent...  but there is a problem on the last part, I will have to make it an assignmet like const something=isOnRightSide ? setRightKnobPosition(...) : setLeftKnob.... but I guess thats not very semantic right?

sahandfarahani:
> The knobs will never drag the first time you click on them. This is because setIsRightKnobDragging(true) is asynchronous, so the mousemove listener will always be using the old value.
> 
Oh!... I didn't notice this... thank you for pointing it!

sahandfarahani:
> The knobs will never drag the first time you click on them. This is because setIsRightKnobDragging(true) is asynchronous, so the mousemove listener will always be using the old value.
> 
> I like how the right knob both serves as the right knob and the single knob when there is only one knob, rather than making a third knob.

Thank you so much ! I didn't notice that!

src/Inputs/Slider.tsx
ryanarine: I would just remove the ternary operator and replace it with the if-else equivalent

src/Inputs/Slider.tsx
ryanarine: Regardless if its a function or not, wouldn't the calculation execute every re-render? (with useMemo the calculation wouldn't execute every re-render but the useMemo itself would)

Does it matter that this is a functional component? if it was class component , the calculation would have to be in the render function(I think) which would execute every re-render (unless you have shouldComponentUpdate)

ralph-dev: @ryanarine Not really because if it was just a function then on a re-render it would just be a function creation not a calculation. UseMemo only helps when someone calls this function to not do the calculation again unless 2 variables changes

src/Inputs/Slider.tsx
```diff
@@ -153,9 +153,12 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
             }
         }
     };
+
     useEffect((): void => {
-        document.addEventListener('mousemove', onMouseMove as any);
-        document.addEventListener('mouseup', onMouseUp);
+        if (isRightKnobDragging === true || isLeftKnobDragging === true) {
```
ryanarine, line 158: can do if (isRightKnobDragging || isLeftKnobDragging)

src/Containers/Popup.tsx
```diff
@@ -0,0 +1,73 @@
...
+    width: 0;
+    height: 0;
+    border-left: 21px solid transparent;
+    border-right: 0px solid transparent;
+    border-top: 24px solid #fff;
+`;
+
+const Container = styled.div<PopupProps>`
+    width: 100px;
+    height: 100px;
```
ryanarine, line 66: Does this need to be here? I removed it and it fixes the overflow problem when the right knob popup is at the end.

ralph-dev:
@indancisive I want you to give the final approval. Look over the code carefully.

src/Containers/Popup.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import styled from 'styled-components';
+import { LabelLayoutProps } from '@Layouts';
+
+export interface PopupProps extends LabelLayoutProps {
+    left?: number;
+    top?: number;
+    width?: number;
+    height?: number;
+    popup?: boolean;
```
indancisive, line 10: does this do 'popup' prop do anything?

src/Inputs/Input.tsx
```diff
@@ -1,12 +1,7 @@
...
-import {
-    LabelLayout,
-    LabelLayoutProps,
-    InputFragment,
-    InputFragmentProps,
-} from '../Fragments';
+import { LabelLayout, LabelLayoutProps, InputFragment } from '@Layouts';
 
-export interface InputProps extends LabelLayoutProps, InputFragmentProps {
+export interface InputProps extends LabelLayoutProps {
```
indancisive, line 4: why did you remove inputfragmentprops ?

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,379 @@
...
+        </LabelLayout>
+    );
+};
+
+const SliderBoard = styled.div<SliderProps>`
         width: 100%;
-        height: 8.4px;
-        cursor: pointer;
+        height: 4px;
         animate: 0.2s;
```
indancisive, line 253: is this correct, i don't think animate is a valid attribute?

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,379 @@
...
+    left: number;
+    right: number;
+    disabled: boolean;
+    hasRail: boolean;
+}
+
+const SliderBoardSelected = styled.div<SelectedBarProps>`
+    top: 1px;
+    height: 4px;
+    animate: 0.2s;
```
indancisive, line 280: here as well, I opened up your branch and it's giving me unknown property, can you just check to make sure this is valid?
image

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,379 @@
...
+         top: -21px;
+`
+            : 'top: -7px; '}
+
+    cursor: grab;
+    ${({ theme }): string => `
+            background: ${theme.colors.primary};
+        `}
+    &:active {
+        box-sizing: border-box
```
indancisive, line 342: missing semi-colon

src/Inputs/Slider.tsx
```diff
@@ -1,79 +1,379 @@
...
+            ? `
+        cursor: not-allowed;
+        opacity: 0.6;
+    `
+            : ' '}
+    ${({ theme }): string => `
+            background: ${theme.colors.primary};
+        `}
+        &:active {
+        box-sizing: border-box
```
indancisive, line 374: missing semi-colon

src/Containers/Popup.tsx
```diff
@@ -0,0 +1,71 @@
...
+    children,
+    left,
+    top,
+    width,
+    height,
+    popup,
+    ...props
+}): React.ReactElement => {
+    return (
+        <Container popup={popup} left={left} top={top}>
```
indancisive, line 23: I don't think pop-up is a standard div attribute can u clarify what this does as it doesn't look like it does anything (I tried passing in false in your story and it just changed nothing)

src/Containers/Popup.tsx
sahandfarahani: It's a boolean saying whether component should have popup or not

src/Inputs/Input.tsx
sahandfarahani: That's strange... I didnt do anything with input.tsx

src/Containers/Popup.tsx
sahandfarahani: Oh it's in popup.tsx
Due to the changes I made to slider.tsx this is useless now thank thank you for pointing it out

PR #98, opened 2020-05-25
Overview Stories

ryanarine:
-Added stories for typography and color
-Added sorting to the stories so Overview shows at the top
-Added ColorCard component similar to the one on the master branch

Typography

Colors


ralph-dev:
@ryanarine Can you update the picture since the UI has been updated

PR #108, opened 2020-07-15
CEM-725-Modify-Image-Carousel-Component

Ampatte2:
Modified the ImageCarousel Component to be able to accept additional props to increase flexibility.

ImageCarousel can accept overlay, overlayText, icon, and pointer as props.

CEM-725-Modify-Image-Carousel

### DIFF
--- a/src/Containers/ImageCarousel.tsx
+++ b/src/Containers/ImageCarousel.tsx
@@ -11,7 +11,11 @@ export interface ImageCarouselProps
         Omit<React.HTMLAttributes<HTMLUListElement>, 'onClick'>,
         ImplicitPropsInterface {
     imageData: string[];
+    pointer:boolean;
     onClick: Function;
+    icon:boolean;
+    overlay:boolean;
+    overlayText:string;
     altText: string;
     width?: number;
     height?: number;
@@ -20,6 +24,10 @@ export interface ImageCarouselProps
 export const ImageCarousel: React.FC<ImageCarouselProps> = ({
     imageData,
     onClick,
+    pointer,
+    icon,
+    overlay,
+    overlayText,
     altText,
     width = 150,
     height = 75,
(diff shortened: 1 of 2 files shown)


src/Containers/ImageCarousel.tsx
```diff
@@ -11,7 +11,11 @@ export interface ImageCarouselProps
         Omit<React.HTMLAttributes<HTMLUListElement>, 'onClick'>,
         ImplicitPropsInterface {
     imageData: string[];
+    pointer:boolean;
```
ralph-dev, line 14:
```suggestion
    pointer: boolean;
```

src/Containers/ImageCarousel.tsx
```diff
@@ -11,7 +11,11 @@ export interface ImageCarouselProps
...
     onClick: Function;
+    icon:boolean;
+    overlay:boolean;
+    overlayText:string;
```
ralph-dev, line 18:
```suggestion
    icon: boolean;
    overlay: boolean;
    overlayText: string;
```

ralph-dev: I would have named it hoverOverlay to make it more clear :)

ralph-dev: or not sure what this does, is it a toggle for isHoverable? in this case the name hasOverlay etc might be better?

src/Containers/ImageCarousel.tsx
```diff
@@ -20,6 +24,10 @@ export interface ImageCarouselProps
 export const ImageCarousel: React.FC<ImageCarouselProps> = ({
     imageData,
     onClick,
+    pointer,
+    icon,
+    overlay,
```
ralph-dev, line 29:
```suggestion
    overlay = true,
```

src/Containers/ImageCarousel.tsx
```diff
@@ -20,6 +24,10 @@ export interface ImageCarouselProps
 export const ImageCarousel: React.FC<ImageCarouselProps> = ({
     imageData,
     onClick,
+    pointer,
+    icon,
```
ralph-dev, line 28:
```suggestion
    icon = true,
```
Not sure why you don't just make this a source and let the person pass an Icon and if they don't pass an Icon then you don't render the Icon ;) This seperate icon is bad practice

ralph-dev: Ex: hoverIcon: source = "X"
hoverIcon -> Clear Name
: source -> Clear Type
"X" -> To not break the existing applciations using this component :)

PR #112, opened 2020-07-23
CEM-757_Ranking_Component

arkiis:
Created styled Ranking Component with sorting and filtering functionality for the total spent. Also added it to storybook and tested with different volumes of data.

screenshot-localhost_6006-2020 07 27-17_28_40

Tablegiftwo

@ralph-dev

### DIFF
--- a/src/Containers/RankingTable.tsx
+++ b/src/Containers/RankingTable.tsx
@@ -12,11 +12,19 @@ interface CustomerProps {
     name: string;
     image: string;
     totalSpent: number;
+    totalSpent1D: number;
+    totalSpent1W: number;
+    totalSpent1M: number;
+    totalSpent1Y?: number;
 }
 interface ItemProps {
     name: string;
     image: string;
     totalSpent: number;
+    totalSpent1D: number;
+    totalSpent1W: number;
+    totalSpent1M: number;
+    totalSpent1Y?: number;
 }
 
 interface RankingTableProps {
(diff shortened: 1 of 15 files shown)


src/Containers/RankingTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+    isAscending: boolean;
+}
+
+const formatter = new Intl.NumberFormat('en-US', {
+    style: 'currency',
+    currency: 'USD',
+});
+
+const useSort = (
+    data: CustomerProps[] | ItemProps[],
```
ralph-dev, line 36: I would call this rowData to give it more semantic meaning.

src/Containers/RankingTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+    rows: number,
```
ralph-dev, line 37: visibleRows
is a more clear name.

ralph-dev: or even rowLimit

src/Containers/RankingTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+        `
+        width: 50px;
+        height: 50px;
+       
+    `,
+    )}
+`;
+const TableHeaderDiv = styled.div`
+    border: 1.5px solid rgba(0, 0, 0, 0.1);
+    background: white;
```
ralph-dev, line 153: Use Theme for colors

src/Containers/RankingTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+        color:${theme.colors.text};
+        font-size:${theme.font.size.default};
+    `}
+`;
+
+const TableRow = styled.tr`
+    display: grid;
+    padding-left: 20px;
+    grid-template-columns: 1fr 3fr 1fr;
+    border-bottom: 1.5px solid rgba(0, 0, 0, 0.1);
```
ralph-dev, line 179: Use Theme for colors

src/Containers/RankingTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+        `
+       grid-template-columns: 1fr 2fr 1fr;
+       
+    `,
+    )}
+`;
+
+const TableBody = styled.tbody`
+    display: block;
+    background: white;
```
ralph-dev, line 193: theme

src/Containers/RankingTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+    border-radius: 0 0 8px 8px;
+    border-right: 1.5px solid rgba(0, 0, 0, 0.1);
+    border-left: 1.5px solid rgba(0, 0, 0, 0.1);
```
ralph-dev, line 196: Short hand notation?

ralph-dev:
Also the image is quite big compared to the Needed Row Height.

src/Containers/RankingTable.tsx
```diff
@@ -73,18 +78,28 @@ export const RankingTable: React.FC<RankingTableProps> = ({
...
-                                                    setIsAscending(!isAscending)
-                                                }
-                                            >
-                                                {heading.toUpperCase()}
-                                            </span>
+                                            <TotalSpentDiv>
+                                                <span
+                                                    role="button"
+                                                    tabIndex={0}
+                                                    onKeyDown={(): void =>
```
ralph-dev, line 85: Why is there two btw?

arkiis: Should I not pass this function to the onKeyDown event?

ralph-dev: Just not sure why you are calling the function twice? isn't onClick enough?

ralph-dev: https://reactnative.dev/docs/pressable
Look at the diagram this might explain things more :)

arkiis: Okay thanks :) but I was getting an eslint error "Visible, non-interactive elements with click handlers must have at least one keyboard listener." [https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/click-events-have-key-events.md](url) for accessibility, but I can ignore the error

ralph-dev: Yes but you have to specify the key that was down to catch the correct key. This is more advanced we can visit later since accessibility is very complex to get right 👍 

For now just ignore the error.

arkiis: Sounds good. Can I merge this into v2 without having the dropdown finished?

ralph-dev: @arkiis Why not just import the existing dropdowns from React-UI?

ralph-dev: UI Layout:
Header: Top 10 Items (HeaderTitle: string as a Prop)
DropDown
Table.

ralph-dev:
You should use the scroll MIXIN to have a nicer scroll experience :)

src/Containers/Settings/AccountCard.tsx
```diff
@@ -38,7 +41,7 @@ export const AccountCard: React.FC<AccountProps> = ({ employee, logout }) => {
             />
             <Button
                 margin="20px 0 0"
-                onClick={() => logout()}
+                onClick={(): void => logout()}
```
ralph-dev, line 44: Just pass logout, since it's just a function reference no need to do 
() => func()
You can do
{logout}
When you are not passing any parameters btw

arkiis: @ralph-dev 

I got a TS error passing the function as a reference so I changed the AccountProps interface from `logout: Function` to `logout: (event: React.MouseEvent<HTMLButtonElement>) => void` 

https://github.com/cheapreats/react-ui-library/blob/v2/src/Containers/Settings/AccountCard.tsx

PR #125, opened 2020-08-13
CEM-855 Added Props

dannybuii:
Added Props

### DIFF
--- a/src/Containers/Settings/InfoCard.tsx
+++ b/src/Containers/Settings/InfoCard.tsx
@@ -26,7 +26,20 @@ export interface InfoProps
     companyPhoneNumberDigits: string;
     techSupportButton: string;
     privacyPolicyButton: string;
+    nonEmergencyLabel: string;
+    assistanceLabel: string;
 }
+
+
+function formatPhoneNumber(companyPhoneNumberDigits) {
+    var cleaned = ('' + companyPhoneNumberDigits).replace(/\D/g, '')
+    var match = cleaned.match(/^(1|)?(\d{3})(\d{3})(\d{4})$/)
+    if (match) {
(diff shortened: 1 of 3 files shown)


src/Containers/Settings/InfoCard.tsx
```diff
@@ -19,12 +19,27 @@ export interface InfoProps
...
+    privacyPolicyButton: string;
 }
 
 export const InfoCard: React.FC<InfoProps> = ({
     faqName,
     faqLink,
     version,
+    companyPrivacyURL = 'https://legal.cheapreats.com/privacy-policy.html',
+    companySupportEmail = 'support@cheapreats.com',
+    companyPhoneNumber = `+1 (306) 341-1035`,
```
ralph-dev, line 37: Having two of the exact same props is not good code practice.

ralph-dev:
Look at how this has Masks to format the phone Number do this as well.

https://github.com/cheapreats/react-ui-library/blob/v2/src/Inputs/MaskedInput.tsx

src/Containers/Settings/InfoCard.tsx
```diff
@@ -63,20 +78,20 @@ export const InfoCard: React.FC<InfoProps> = ({
...
                     {'For non-emergency inquiries, shoot us an email at '}
-                    <a href="mailto:support@cheapreats.com">
-                        support@cheapreats.com
+                    <a href={`mailto:${companySupportEmail}`}>
+                        {companySupportEmail}
                     </a>
                     .
                 </Paragraph>
                 <Paragraph margin="15px 0" bold>
                     {'If you require immediate assistance, please call/text: '}
```
ralph-dev, line 91: What about this as a prop? 
Assistance_Label etc

PR #131, opened 2020-08-17
CEM-861-Fix-Switch-Component

Ampatte2:
Modified the switch component to be useable/readable. 

This will break the existing switch useage in EmployeeDetails.tsx and SettingsSwitch in the Terminal.

CEM-861-Fix-Switch-Switch-Color
CEM-861-Fix-Switch-Active-Color
CEM-861-Fix-Switch-Labels
CEM-861-Fix-Switch-Default

### DIFF
--- a/src/Inputs/Switch.tsx
+++ b/src/Inputs/Switch.tsx
@@ -1,133 +1,127 @@
 import React from 'react';
 import styled from 'styled-components';
 import { position, darken, flex, transition } from '../Utils/Mixins';
-import { LabelLayout, LabelLayoutProps } from '../Fragments';
 
-const COMPONENT = (
-    props = {},
-    innerProps = {},
-    value?: boolean,
-): React.ReactElement => (
-    <Container>
-        <Input
-            type="checkbox"
-            {...props}
-            value={value === undefined ? '' : value.toString()}
-            checked={value}
-        />
-        <SwitchBox {...innerProps}>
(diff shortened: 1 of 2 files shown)


src/Inputs/Switch.tsx
```diff
@@ -1,133 +1,127 @@
...
+        <Layout>
+            <Label>{label}</Label>
+            <Info>{description}</Info>
+        </Layout>
+        <Row>
+            {leftTag && <Tag margin="5px 10px 0 0">{leftTag}</Tag>}
+            <Container>
+                <Input
+                    type="checkbox"
+                    onChange={(): void => onChange()}
```
ralph-dev, line 42: Will you not be exposing the value?
Ex: https://reactnative.dev/docs/switch#onchange

src/Inputs/Switch.tsx
```diff
@@ -1,133 +1,127 @@
...
+        <Layout>
+            <Label>{label}</Label>
+            <Info>{description}</Info>
+        </Layout>
+        <Row>
+            {leftTag && <Tag margin="5px 10px 0 0">{leftTag}</Tag>}
+            <Container>
+                <Input
+                    type="checkbox"
+                    onChange={(e): void => onChange(e)}
```
fpunny, line 42: only issue i have with this is that it wouldnt be .value, but .checked which would have to be implemented everywhere else

PR #143, opened 2020-09-07
CEM-937 Store Hours Sets UI

xsharonhe:
[UPDATED]: Incorporated Andrew's feedback in regards to error handling of adding more than one time. I also disabled the button if none of the boxes were checked:
error-more-than-one

[UPDATED]: Time row component that Ralph suggested.
time-row-component

[UPDATED]: All components have been separated. Functionality works the same.
updated-component

[UPDATED]: I added error handling similar to the one from flash sales and used Andrew's suggestions to prevent repeated categories or empty categories from being added.

[UPDATED]: Added the isActive boolean so that it does not depend on local state.
isActive-state-changes

This is it working on a larger width screen (tablet):
tablet-storehourslist-view

This is it working on a smaller width screen (small phone) and shows some of the features I will speak to later on:
smallphone-storehourslist-view

There are some caveats to the code I implemented, so just wanted this to be a heads up, and to check whether any of these will pose as a problem.

- You cannot delete an 'active' category which is the category of times that will be shown to the users. I thought this would be the easiest solution to implement to prevent errors.
- There is another error that I didn't exactly show but you are not allowed to have no categories. You must have at least one category at all times.
- You can only show one time slot in each given day.


src/Containers/StoreHoursList/Alert.tsx
```diff
@@ -0,0 +1,47 @@
...
+        ResponsiveInterface,
+        React.HTMLAttributes<HTMLSpanElement> {
+            text?: string,
+            icon?: React.ForwardRefExoticComponent<React.RefAttributes<SVGSVGElement>>,
+            success?: boolean,
+            error?: boolean
+};
+
+export const Alert: React.FC<AlertProps> = ({
+    text = 'Some text here',
```
ralph-dev, line 19: Shouldn't this be a required prop? With no default parameter?

src/Containers/StoreHoursList/constants.ts
```diff
@@ -0,0 +1,146 @@
+interface DICT {
+    [key: string]: {
+        [key: string]: string; 
+    };
+}
+
+interface DICTINDEXES {
```
ralph-dev, line 7: IDICT_INDEXES
Always start interfaces with I

src/Containers/StoreHoursList/constants.ts
```diff
@@ -0,0 +1,146 @@
...
+        tuesday: { to: string, from: string }[],
+        wednesday: { to: string, from: string }[],
+        thursday: { to: string, from: string }[],
+        friday: { to: string, from: string }[],
+        saturday: { to: string, from: string }[],
+        sunday: { to: string, from: string }[]
+    }
+};
+
+export const sampleCategories: CategoryWithHoursTypes [] = [
```
ralph-dev, line 60: This should be in the Storybook, not the component becuase this is only used for testing

stories/Containers/StoreHoursList.stories.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { StoreHoursList } from '../../src';
+
+storiesOf('StoreHoursList', module)
+    .add('with default', () => (
+        <StoreHoursList/>
```
ralph-dev, line 7: How will this component remember which category was selected as active? How will this component know which data it contains?

The UI looks great, it's on the right track, the code structure has some design issues that can easily be fixed as you start sorting your code in the correct place :)

ralph-dev (changes requested):
Some design issues, let's fix these comments first, then see if you have fixed them otherwise we can jump on a call. I left some thought provoking comments that should spark the solution for you :)

src/Containers/StoreHoursList/Alert.tsx
xsharonhe: I was not able to pass it as a required prop without it showing me errors (without a default parameter). Here's the log: 
image
I passed it as children instead to fix the type error.

src/Containers/StoreHoursList/constants.ts
```diff
@@ -0,0 +1,58 @@
...
+    [key: string]: {
+        [key: string]: string; 
+    };
+}
+
+interface IDICTINDEXES {
+    [key: string]: number
+};
+
+export const constants: IDICT = {
```
ralph-dev, line 11: This should be in the storybook file becuase this is only used there.

src/Containers/StoreHoursList/constants.ts
```diff
@@ -0,0 +1,58 @@
+interface IDICT {
+    [key: string]: {
+        [key: string]: string; 
+    };
+}
+
+interface IDICTINDEXES {
```
ralph-dev, line 7: Put _ around words when using all caps.
I_DICT_INDEXES,

stories/Containers/StoreHoursList.stories.js
```diff
@@ -0,0 +1,99 @@
...
+        saturday: [], 
+        sunday: [] 
+    }
+};
+
+storiesOf('StoreHoursList', module)
+    .add('with default', () => (
+        <StoreHoursList
+            allCategories={sampleCategories}
+            oneCategorySchedule={sampleActiveCategorySchedule}
```
ralph-dev, line 97: What does this prop mean versus the other prop?
Who is defining the active schedule is it the data structure or is it a prop called activeCategorySchedule?

ralph-dev (changes requested):
Looking better, make sure to update your PR with any enhancements you make so that older comments are tended to

ralph-dev:
@xsharonhe 
Go to terminal.CheaprEats.com and login with the test credentials provided in onboarding and try creating a flash sale with the time being incorrect. The UX is much better.


https://github.com/cheapreats/cheapreats-terminal/blob/master/src/pages/Dashboard/Inventory/FlashSale/FlashSaleForm.tsx
Or look at code here ^

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,550 @@
...
+import { BusinessTime } from '@styled-icons/fa-solid/BusinessTime';
+import { Edit } from '@styled-icons/boxicons-regular/Edit';
+import { Add } from '@styled-icons/ionicons-outline/Add';
+import { CircleWithCross } from '@styled-icons/entypo/CircleWithCross';
+import { Alert } from './Alert';
+import { SettingsCard } from '../SettingsCard';
+import { Modal } from '../Modal';
+import { Tag } from '../Tag';
+import { Heading, SmallText } from '../../Text';
+// eslint-disable-next-line import/no-cycle
```
ralph-dev, line 12: You shouldn't have a cycle issue

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,550 @@
...
+        if (cannotDeleteActiveError) {
+            return (
+                <Alert error icon={CircleWithCross}> 
+                    {' '}
+                    { constants.ERRORS.CANNOT_DELETE_ACTIVE_CATEGORY }
+                    {' '}
+                </Alert> 
+            );
+        }
+        if (onlyOneTimeError) {
```
ralph-dev, line 154: These if statements should just be abstracted into a switch statement inside a function that returns alert so no need to define alert 2 times below

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,550 @@
...
+        return null;
+    };
+
+    const convertTime = (date: string, toggle: boolean): string => {
+        if (!toggle) {
+            const d = parseInt(
+                date.slice(
+                    START_INDEX_OF_SELECTION,
+                    END_INDEX_OF_SELECTION_FOR_CONST_D,
+                ), 10
```
ralph-dev, line 177: 10 is a magic number what does it mean?

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,550 @@
...
+                    <StyledButton onClick={(): void => setIs24(!is24)}> 
+                        { constants.BUTTONS.TOGGLE } 
+                    </StyledButton>
+                </ButtonsContainer>
+                <StyledHeading type='h6'>
+                    { constants.TITLES.OPERATIONS }
+                    { activeCategorySchedule.category }
+                </StyledHeading>
+                {Object.entries(activeCategorySchedule.hoursByDay).map((day): React.ReactElement | null => {
+                    const capitalDay = day[0].replace(/^\w/, (chr: string): string =>
```
ralph-dev, line 266: Regex is magic string, define a const MATCH_X_PATTERN
Give X a meaningful name

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,550 @@
...
+                <DaysDiv>
+                    <StyledHeading type='h6'> 
+                        { constants.TITLES.SELECT_A_DAY } 
+                    </StyledHeading>
+                    {Object.entries(checkboxes).map((checked): React.ReactElement => {
+                        return (
+                            <StyledCheckbox
+                                key={checked[CHECKED_INITIAL_INDEX]}
+                                label={checked[CHECKED_INITIAL_INDEX].replace(
+                                    /^\w/,
```
ralph-dev, line 427: Nit

ralph-dev (changes requested):
Code quality needs improvement as well as the UX for time selection

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,550 @@
...
+            allCategories: ICategoryWithHoursTypes[],
+            constants: I_DICT
+};
+
+export const StoreHoursList: React.FC<StoreHoursListProps> = ({
+    allCategories,
+    constants
+}): React.ReactElement => {
+    const editModal = useState(false);
+    const [editModalState, setEditModalState] = editModal
```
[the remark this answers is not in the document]
xsharonhe: I get the following error when I do this:
image
I think it might be because the state is on when you click a button and thus that makes the state 'iterable'.

src/Containers/StoreHoursList/Alert.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import styled from 'styled-components';
+import { CheckCircle } from '@styled-icons/boxicons-regular/CheckCircle';
+import { Card } from '../Card';
+import { MainInterface, ResponsiveInterface } from '../../Utils/BaseStyles';
+import { styledCondition } from '../../Utils/Mixins';
+
+interface AlertProps extends 
+    MainInterface,
```
junzhengca, line 9: Why new line here? please format this interface properly.

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,529 @@
...
+        tuesday: { to: string, from: string }[],
+        wednesday: { to: string, from: string }[],
+        thursday: { to: string, from: string }[],
+        friday: { to: string, from: string }[],
+        saturday: { to: string, from: string }[],
+        sunday: { to: string, from: string }[]
+    }
+};
+
+interface I_DICT {
```
junzhengca, line 33: Seems to be pretty common, maybe move this to a generic file?

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,529 @@
...
+        [key: string]: string; 
+    };
+}
+
+interface ITimeTypes {
+    to: Date | string 
+    from: Date | string 
+};
+
+interface StoreHoursListProps
```
junzhengca, line 44: Formatting is weird.

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,529 @@
...
+const FIRST_TIME = 0;
+const ALL_CATEGORIES_INDEX = 0;
+const ALL_CATEGORIES_TIMES = 1;
+const FIRST_CATEGORY = 0;
+const RADIX_BASE_TEN = 10;
+const MATCH_FIRST_LETTER_PATTERN = /^\w/;
+
+export const StoreHoursList: React.FC<StoreHoursListProps> = ({
+    allCategories,
+    constants
```
junzhengca, line 71: What is `constants`? Make the name more informative or add documentation.

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,529 @@
...
+    const [selectActiveCategory, setSelectActiveCategory] = useState(allCategoriesWithHours[FIRST_CATEGORY].category);
+    const [addStoreHoursCategory, setAddStoreHoursCategory] = useState(allCategoriesWithHours[FIRST_CATEGORY].category);
+    const [activeCategorySchedule, setActiveCategorySchedule] = useState<ICategoryWithHoursTypes>(allCategoriesWithHours[FIRST_CATEGORY]);
+
+    const [success, setSuccess] = useState(false);
+    const [error, setError] = useState('');
+
+    const [is24, setIs24] = useState(true);
+
+    const createCategoryWithHours = (categoryName: string): ICategoryWithHoursTypes => {
```
junzhengca, line 109: violation of JS style guide, all functions must have JSDoc.

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,529 @@
...
+            );
+            date = `${d % FIRST_HALF_OF_A_DAY_IN_HOURS ||
+                FIRST_HALF_OF_A_DAY_IN_HOURS}:${date.slice(
+                START_INDEX_OF_SELECTION_FOR_DATE,
+            )} ${d < FIRST_HALF_OF_A_DAY_IN_HOURS ? 'AM' : 'PM'}`;
+        }
+        return date;
+    };
+
+    const convertDateToHours = (timeObj: ITimeTypes): ITimeTypes | null => {
```
junzhengca, line 177: These functions can be moved to another file and you can do unit tests.

junzhengca (changes requested):

ralph-dev (changes requested):

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,529 @@
...
+                            setEditModalState(!editModalState);
+                        }}
+                    >
+                        { constants.BUTTONS.EDIT }
+                    </StyledButton>
+                    <StyledButton onClick={(): void => setIs24(!is24)}> 
+                        { constants.BUTTONS.TOGGLE } 
+                    </StyledButton>
+                </ButtonsContainer>
+                <StyledHeading type='h6'>
```
fpunny, line 249: Is this h6 hierachically or are you doing it for just the size?

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,529 @@
...
+                            </StyledTag>
+                        </StoreHoursContainer>
+                    ) : (
+                        <Heading bold size='0.95em' margin='0'>
+                            { capitalDay }
+                        </Heading>
+                    )
+                })}
+            </SettingsCard>
+            <StyledModal state={editModal}>
```
fpunny, line 285: This should be a component in itself

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,529 @@
...
+                    </StyledButton>
+                    <StyledButton onClick={(): void => setIs24(!is24)}> 
+                        { constants.BUTTONS.TOGGLE } 
+                    </StyledButton>
+                </ButtonsContainer>
+                <StyledHeading type='h6'>
+                    { constants.TITLES.OPERATIONS }
+                    { activeCategorySchedule.category }
+                </StyledHeading>
+                {Object.entries(activeCategorySchedule.hoursByDay).map((day): React.ReactElement | null => {
```
fpunny, line 253: This can be a component in itself to help keep the bulk down here

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,529 @@
...
+                            setSuccess(false);
+                            setError('');
+                        }}
+                    > 
+                        { constants.BUTTONS.EDIT_CATEGORIES }
+                    </StyledButton>
+                </ButtonsContainer>
+                { showError(error) }
+                { showSuccess(constants.SUCCESS.CATEGORY_CREATED) }
+                <StyledModal state={editCategoryModal}>
```
fpunny, line 313: This should be a component in itself

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,529 @@
...
+                        onClick={(): void => {
+                            setActiveCategorySchedule(getActiveSchedule(selectActiveCategory));
+                            setError('');
+                        }}
+                    > 
+                        { constants.BUTTONS.SET_ACTIVE }
+                    </StyledButton>
+                </ButtonsContainer>        
+            </StyledModal>
+            <StyledModal state={addModal}>
```
fpunny, line 394: This should be a component in itself

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,529 @@
...
+                    }}
+                >
+                    { constants.BUTTONS.ADD_HOURS } 
+                </CenteredButton> 
+            </StyledModal>
+        </StoreHoursContainer>
+    );
+};
+
+const StoreHoursContainer = styled.div`
```
fpunny, line 473: just use a div

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,529 @@
...
+`;
+const TagContainer = styled.div`
+    margin: auto;
+    ${({ theme }): string => `
+        padding: ${theme.dimensions.padding.container};
+    `};
+`;
+const StyledTag = styled(Tag)`
+    margin: 5px;
+`;
```
fpunny, line 529: 3 things with exactly the same style implies there should be relationship? maybe they are just spacing for sections of a component?

xsharonhe: Hi Frederic, I did notice that as well. They are spacing for sections of the component. What would be the better way to make this code less redundant?

src/Containers/StoreHoursList/StoreHoursList.tsx
xsharonhe: This would be just for the size.

src/Containers/StoreHoursList/StoreHoursList.tsx
fpunny: Make a SC called Section, and change the type using as
```
<Section as={OWO}>
...
</Section>

...

const Section = styled.div`margin: 5px`;
```

src/Containers/StoreHoursList/StoreHoursList.tsx
fpunny: use CSS, html is for structure and CSS is for styling.

src/Containers/StoreHoursList/StoreHoursList.tsx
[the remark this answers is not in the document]
fpunny: > Why? Just do it on the same line to reduce code

modal takes state as itself, so u need to deconstruct/index the array to use the state/dispatch

src/Containers/StoreHoursList/StoreHoursList.tsx
xsharonhe: This is part of the Heading UI component as a prop. Would you still recommend to separate the two pieces of logic.

For reference this is the written Heading component:
```
enum elements {
    h1 = 'h1',
    h2 = 'h2',
    h3 = 'h3',
    h4 = 'h4',
    h5 = 'h5',
    h6 = 'h6',
}

export const Heading: React.FC<HeadingProps> = ({
    children,
    type = 'h1',
    size = elements[type] || elements.h1,
    ...props
}): React.ReactElement => (
    <TextLayout type={type} size={size} {...props}>
        {children}
    </TextLayout>
);

export default Heading;
```

fpunny: Hmm in that case, change the size... I wish I can remake the ui library again, but im not an active member anymore, and ralph would prob strangle me

src/Containers/StoreHoursList/StoreHoursList.tsx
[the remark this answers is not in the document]
xsharonhe: Could you provide a sample code of how to do this? I've never seen useState being restructured before.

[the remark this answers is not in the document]
fpunny: > Could you provide a sample code of how to do this? I've never seen useState being restructured before.

no wat ur doing is fine, theres nothing wrong with it. We cant reconstruct it or else React would detect a new array every re-render, which would fk the performance of the app 100%

src/Containers/StoreHoursList/StoreHoursList.tsx
xsharonhe: Would this category not rely on too many other states to make into a separate component? (e.g. is24 toggle on/off to make it AM/PM and 24 hours) and setting the active Schedule. Or would you put these as props?

ralph-dev:
@xsharonhe Look at how Flash Sale implements TO and FROM. Your error handling has bad UX compared to that one. I mentioned dit earlier

Ampatte2:
This part of the component has some issues. If I build a category and take time to add hours there should be a confirmation before I can delete it, feels like it could create frustration.

Also I can add a blank category. Should have a minimum amount of text to create a category.

CEM-937-Store-Hours Empty Category

src/Utils/Constants/dict.ts
```diff
@@ -0,0 +1,5 @@
+export interface I_DICT {
+    [key: string]: {
+        [key: string]: string; 
+    };
+};
```
Ampatte2, line 5: Maybe Fred or Ralph will correct me, but I think this should be typed how you are using it in the component. The component doesn't utilize variable I_DICT props so I don't see why you wouldn't strongly type the props to match how you are using them.

src/Containers/StoreHoursList/StoreHoursList.tsx
ralph-dev: @fpunny Isn't this already a Modal component, it's just styling being applied? Or are you saying to make this a component with the styling preset?

src/Containers/StoreHoursList/TimeFunctions.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const from = parsedFromDate.toLocaleTimeString('it-IT')
+        .slice(START_INDEX_OF_SELECTION, END_INDEX_OF_SELECTION_DATE_TO_HOURS);
+    if (parsedFromDate >= parsedToDate) {
+        return null;
+    } 
+    timeObj = {
+        'to': to,
+        'from': from
+    };
+    return timeObj;
```
Ampatte2, line 50: Utils has a FormatDate function might be helpful here.

src/Utils/Constants/dict.ts
ralph-dev: I believe the typing for this can be a bit more strict with the keys Sun - Saturday being explicitly defined but because you have an 0-infinite length of categories it will be impossible to type the root level.

src/Containers/StoreHoursList/StoreHoursList.tsx
```diff
@@ -0,0 +1,508 @@
...
+const CATEGORY_SCHEDULE = 1;
+const MATCH_FIRST_LETTER_PATTERN = /^\w/;
+
+export const StoreHoursList: React.FC<StoreHoursListProps> = ({
+    allCategories,
+    textHeaders
+}): React.ReactElement => {
+    
+    const editModal = useState(false);
+    const [editModalState, setEditModalState] = editModal;
```
ralph-dev, line 48: just use

const [isEditModalVisible, setIsEditModalVisible] = useState(false);

src/Containers/StoreHoursList/ConfirmModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+    isVisible: [boolean, React.Dispatch<React.SetStateAction<boolean>>],
+    confirmDelete: string,
+    yesButtonLabel: string,
+    noButtonLabel: string,
+    allCategories: ICategoryWithHoursTypes[],
+    setAllCategories: React.Dispatch<React.SetStateAction<ICategoryWithHoursTypes[]>>
+    deletedCategory: string
+};
+
+export const ConfirmModal: React.FC<ConfirmModalProps> = ({
```
ralph-dev, line 22: Just make sure to always pass ...props for extensibility!

ralph-dev (commented):
Make sure ...props everywhere

ralph-dev:
@xsharonhe UX update, I would put the From and To Side by Side like Fred did in the Terminal Image above :)
This way when you click the time, it doesn't block the other time

ralph-dev:
> @xsharonhe UX update, I would put the From and To Side by Side like Fred did in the Terminal Image above :)
> This way when you click the time, it doesn't block the other time

I would also make this a component so other people don't make this mistake

ralph-dev:
I would call the component
FromToDualTimeSelector? Its more clear but otherwise your code looks much better!

ralph-dev (approved):
LGTM
@fpunny  thoughts?

Ampatte2:
Fred, Ralph would this be better named as toggle 24hrs?

CEM-937 Toggle AM-PM

xsharonhe:
> This confuses me, it gives me an error if I select more than one day, but it also updates the hours for more than one day. Error message also closes the Add Store Hours modal so I have to re-open if I make an error, if it is an error?
> 
> CEM-937-Add Store Hours

It updates the hours for those days that don't have more than one day. For example, if Monday already has at time, but you select Monday, Tuesday, Wednesday, it will update for Tuesday and Wednesday. But I see how that affects the user's experience, so I'll make the necessary changes.

src/Containers/StoreHoursList/CreateHoursModal.tsx
```diff
@@ -33,6 +33,16 @@ const CHECKBOX_DAY = 0;
 const CHECKBOX_TIME = 1;
 const MATCH_FIRST_LETTER_PATTERN = /^\w/;
 
+const initialCheckboxState = {
```
ralph-dev, line 36: I would create a Typescript interface for this btw

src/Containers/StoreHoursList/CreateHoursModal.tsx
```diff
@@ -70,6 +71,28 @@ export const CreateHoursModal: React.FC<CreateHoursProps> = ({
 
     const [error, setError] = useState('');
 
+    useEffect((): void => {
+        setCheckboxes(initialCheckboxState);
+    }, [addModalState])
+
+    /**
+     * Checks if there is more than one time
```
ralph-dev, line 79: time?
This naming is quite confusing...

PR #144, opened 2020-09-08
CEM-947-Fix-Typescript-Errors-Orders-Migration

Ampatte2:
The orders migration to typescript brought up typing errors with Card since it did not extend main and responsive interfaces.

KitchenCard and KitchenCardItems were incorrectly typed as [ Item ] when it is supposed to be an array of items, Item[ ].

SettingsSwitch is a component in the terminal, adding to library.


ralph-dev (approved):
Lgtm

PR #146, opened 2020-09-09
CEM-949-Settings-Card-Component

Ampatte2:
The terminal has this as a component even though it is in the library. Updated SettingsCard to have onClick mouseEvent and added conditional cursor when icon is clickable.

Removed unneccessary import ImageCarouselStories.


ralph-dev (approved):
LGTM

PR #149, opened 2020-09-14
CEM-971 Content Component

Ampatte2:
Adding content component from terminal to library.

There are issues with babel 7.8.6 as seen here https://github.com/storybookjs/storybook/issues/10477 . Upgrading to babel 7.9.0.

Should this be a layout?

CEM-971 Content Component

### DIFF
--- a/src/Containers/Content.tsx
+++ b/src/Containers/Content.tsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import styled, { DefaultTheme, InterpolationFunction } from 'styled-components';
+import { scroll, media } from '../Utils/Mixins';
+
+export interface ContentProps extends React.HTMLAttributes<HTMLDivElement> {
+    containerProps?: React.HTMLAttributes<HTMLDivElement>;
+    withList?: boolean;
+}
+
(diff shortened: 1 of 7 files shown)


src/Containers/Content.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import styled, { DefaultTheme, InterpolationFunction } from 'styled-components';
+import { scroll, media } from '../Utils/Mixins';
+
+export interface ContentProps extends React.HTMLAttributes<HTMLDivElement> {
+    containerProps?: React.HTMLAttributes<HTMLDivElement>;
+    withList?: boolean;
+}
+
+export const Content: React.FC<ContentProps> = ({
```
ralph-dev, line 10: Is this scrollableListContent?

ralph-dev (approved):
Lgtm I guess, missing withList = true case

Ampatte2:
That makes more sense. What do you mean by withList true case, in the storybook?

ralph-dev:
> That makes more sense. What do you mean by withList true case, in the storybook?

Nvm ignore that comment, I saw you added a Knob which is perfect

PR #151, opened 2020-09-15
CEM-978 DisplayItem Optional Props

Ampatte2:
In order to accept current terminal implementations of DisplayItem the props are being changed to optional.


ralph-dev (approved):
LGTM

PR #153, opened 2020-09-18
Upgraded Storybook to v6.0 and organized it properly

ralph-dev:
Upgrade to Storybook V6
Migrate 1 Story to TypeScript
Format things better 

Upgrade every dependency in the package.json


stories/Themes/index.stories.tsx
```diff
@@ -2,8 +2,9 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { withKnobs } from '@storybook/addon-knobs';
 import * as Themes from '../../src/Themes';
+import {createStoryTitle} from "../Constants";
 
-const stories = storiesOf('Theme', module).addDecorator(withKnobs);
+const stories = storiesOf(createStoryTitle('Theme'), module).addDecorator(withKnobs);
 Object.entries(Themes).forEach(([key, value]) => {
     stories.add(`with ${key}`, () => <span>To be automated</span>);
 });
```
Ampatte2, line 10: This story's purpose is to show light or dark theme? With the additon of Design system colors this seems redundant, but maybe it will be used to display components?

Ampatte2 (changes requested):
TextArea and TimePicker are both included in ClientShowCase story. If this is intentional Timepicker and TextArea stories should have a label of the component they are displaying.

.storybook/main.js
```diff
@@ -0,0 +1,29 @@
+const { aliases } = require('../package.json');
+const path = require('path');
+
+module.exports = {
+    stories: ['../stories/**/*.stories.@(js|mdx|tsx|ts|jsx)'],
+    addons: ['@storybook/addon-docs', '@storybook/addon-viewport', '@storybook/addon-knobs/', '@storybook/addon-links/', '@storybook/addon-actions/', '@storybook/addon-toolbars', '@storybook/addon-a11y', '@storybook/addon-controls', '@storybook/addon-backgrounds'],
+    webpackFinal: async (config, { configType }) => {
```
fpunny, line 7: I believe storybook has a recipe for this now, should look into that

.storybook/preview.js
```diff
@@ -0,0 +1,87 @@
...
+            method: 'alphabetical',
+            order: ['Design System'],
+            locales: 'en-US',
+        },
+    },
+    backgrounds: {
+        default: '',
+        values: [
+            {
+                name: 'CheaprEats White',
```
fpunny, line 35: should grab these values from theme

package.json
```diff
@@ -15,7 +15,7 @@
         "prepublish": "npm run lint && npm run build",
         "eslint": "eslint ./src --ext tsx,ts,jsx,js",
         "eslint:fix": "npm run eslint -- --fix",
-        "prettier": "prettier **/*.ts **/*.tsx **/*.js **/*.jsx",
+        "prettier": "prettier **/*.ts **/*.tsx **/*.js",
```
fpunny, line 18: include the storybook files cus they need formatting

package.json
```diff
@@ -36,36 +36,36 @@
         "styled-icons": "*"
     },
     "devDependencies": {
```
fpunny, line 38: the deps need to be reorganized for proper ci/cd in the future

src/Inputs/MaskedInput.tsx
```diff
@@ -39,17 +39,14 @@ const PERCENT_FORMAT_MASK = (s: string): string => {
...
-            default:
-                return (): string => 'Invalid mask.';
-        }
+    switch (mask) {
+        case MaskedInputPreset.DOLLAR:
+            return DOLLAR_FORMAT_MASK;
+        case MaskedInputPreset.PERCENTAGE:
+            return PERCENT_FORMAT_MASK;
+        default:
+            return mask as (value: string) => string;
```
fpunny, line 48: This is not a good practice... there should be this check which is overwritten depending on the node environment to keep developers in check without exploding in production

src/Inputs/Slider.tsx
```diff
@@ -248,23 +248,23 @@ export const Slider: React.FunctionComponent<SliderProps> = ({
...
 
 const SliderBoard = styled.div<SliderProps>`
-        width: 100%;
-        height: 4px;
-        background: #e9e9e9;
-        cursor:pointer;
-        ${({ theme }): string => `
+    width: 100%;
+    height: 4px;
+    background: #e9e9e9;
```
fpunny, line 253: rip theming

stories/Constants.ts
```diff
@@ -0,0 +1,22 @@
+import {useGlobals} from "@storybook/addons";
+
+export const COMPONENTS_TITLE = 'Components';
+
+export const createStoryTitle = (componentName: string) => {
```
fpunny, line 5: fairly general name for a specific action

.storybook/main.js
ralph-dev: I think your referring to essentials?

.storybook/preview.js
ralph-dev: Good idea

src/Inputs/MaskedInput.tsx
ralph-dev: Not sure what you mean here, all I did was remove the If statement

stories/Constants.ts
ralph-dev: Just wanted to abstract it away incase we need to change in the future is easy (looks a bit overkill for now I agree but worth it in long run lol)

.storybook/main.js
fpunny: nvm they got rid of it in this major version...

fpunny: nvm i found it, its a preset mb
https://www.npmjs.com/package/@storybook/preset-typescript

fpunny: maybe storybook is more "developed", but i have a feeling the build stuff would impode for some reason

ralph-dev: @fpunny Storybook comes with TypeScript configuration built in now? Is that still needed?

stories/Themes/index.stories.tsx
ralph-dev: @fpunny what is this file used for? Can't we just use Storybook Globals to change the theme?

src/Inputs/MaskedInput.tsx
fpunny: The if statement is in place to ensure that developers using this component don't use it incorrectly and let them know that its used incorrectly. So im saying u shouldnt remove it if the code is in a dev environment, but remove it when the code is in production... which can be done using NODE_ENV

ralph-dev: But typescript will do the check for you so not sure of the point?

PR #154, opened 2020-09-19
Converted Card Component

eric-prog:
image


ralph-dev (approved):
Lgtm, let's try to get all the components before monday if possible

PR #155, opened 2020-09-22
CEM-1012 List Component

Ampatte2:
List Component from terminal added to React-UI library.

Typescript error for LabelLayout in layouts and fragments gave error of circularly referencing itself. It can be solved by not referencing, upgrading typescript and styled-components version (that will be done in the storybook upgrade), or styled(Label as any). Choose not to reference.


src/Containers/List/List.tsx
```diff
@@ -0,0 +1,115 @@
+import React, { useState, useEffect } from 'react';
+import styled, { DefaultTheme, InterpolationFunction } from 'styled-components';
+import { Mixins } from '../../Utils';
+import { Responsive, Main } from '../../Utils/BaseStyles';
+import { Loading } from '../Loading';
+import { ListStateContext, ListDispatchContext } from './ListContext';
+import { ListToggle } from './ListToggle';
+
+export interface ListProps {
```
ralph-dev, line 9: does this not extend div since the ...props are on container?

ralph-dev (approved):
LGTM, I will merge the V2 Upgrade now becuase these are using the older format

ralph-dev:
image

Notice that the left side bar here has a Collapsible Arrow option. Wondering if should support this functionality here or make a different component called leftSideBar?

PR #156, opened 2020-09-22
Store Feature Card

eric-prog:
image


ralph-dev (approved):
LGTM, your missing a few controls but the general concept is correct

ralph-dev:
Why did you remove tags?

ralph-dev (approved):
Lgtm, but have the props inside args when possible! Ideally tags should be in props so you can change them in controls somehow but that's a problem for another day

PR #161, opened 2020-09-23
CEM-1016 Typescript Errors React-UI

Ampatte2:
Upgrading typescript version and parser caused default rules to be updated as well.

Updated eslintrc.js with rules.

### DIFF
--- a/src/Containers/ImageCarousel.tsx
+++ b/src/Containers/ImageCarousel.tsx
@@ -25,7 +25,9 @@ export interface ImageCarouselProps
 
 export const ImageCarousel: React.FC<ImageCarouselProps> = ({
     imageData,
-    onClick = (): void => {},
+    onClick = (): void => {
+        return undefined;
+    },
     pointer = true,
     hoverIcon = Times,
     hoverOverlay = true,
--- a/src/Inputs/MaskedInput.tsx
+++ b/src/Inputs/MaskedInput.tsx
@@ -1,14 +1,17 @@
 import React, { useState, useEffect } from 'react';
-import { LabelLayout, LabelLayoutProps, InputFragment } from '../Fragments';
+import {
+    LabelLayout,
+    LabelLayoutProps,
+    InputFragment,
+    InputFragmentProps,
+} from '../Fragments';
 
 export enum MaskedInputPreset {
     DOLLAR = 'DOLLAR',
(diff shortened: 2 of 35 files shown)


ralph-dev (commented):
LGTM

Does that need return undefined? I'm not sure void === undefined? For the function part though

src/Containers/ImageCarousel.tsx
```diff
@@ -25,7 +25,9 @@ export interface ImageCarouselProps
 
 export const ImageCarousel: React.FC<ImageCarouselProps> = ({
     imageData,
-    onClick = (): void => {},
+    onClick = (): void => {
+        return undefined;
```
ralph-dev, line 29: Weird

src/Inputs/MaskedInput.tsx
```diff
@@ -1,14 +1,17 @@
...
 
 export enum MaskedInputPreset {
     DOLLAR = 'DOLLAR',
     PERCENTAGE = 'PERCENTAGE',
 }
 
-export interface MaskedInputProps
-    extends LabelLayoutProps,
-        React.InputHTMLAttributes<HTMLInputElement> {
+export interface MaskedInputProps extends LabelLayoutProps, InputFragmentProps {
```
ralph-dev, line 14: No need for HTML? There is a props spread, is that covered by the others?

ralph-dev (approved):
Lgtm

src/Inputs/MaskedInput.tsx
Ampatte2: InputFragmentProps already extends HTMLInputElement.

src/Containers/ImageCarousel.tsx
Ampatte2: Either turn off no empty functions, or this is the correct way to type a void function in typescript.

Ampatte2:
Typing wise void === undefined.

src/Containers/ImageCarousel.tsx
ralph-dev: Let's turn off no empty functions then @fpunny

Ampatte2:
I do not know why but removing quiet fixes the checks and build issues.

ralph-dev (approved):
LGTM, Good catch!

PR #171, opened 2020-09-28
ClientShowCase

eric-prog:
image


ralph-dev (commented):
Select is imported from ADD ON KNOBS which means its a special data type that just wants to be represented an ENUMs this means you can remove Select and with Controls just do Arg Typ, ENUM 👍

stories/Containers/ClientShowCase.stories.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import styled from 'styled-components';
 import { Button } from '../../src/Inputs/Button';
+import { select } from '@storybook/addon-knobs';
```
ralph-dev, line 4: DO NOT USE ADDON KNOBS do not import this. Ths

ralph-dev: https://storybook.js.org/docs/react/api/argtypes#gatsby-focus-wrapper
read this

ralph-dev (approved):
LGTM

PR #183, opened 2020-09-30
CEM-1026 Modify KitchenCard to accept choice_type format

Ampatte2:
Modified KitchenCard to use the new GraphQL modifier type of choice_type.

Updated Status component and Status component's implementation to use new statusColors which require filtering.

Status enum was updated on GraphQL to be all uppercase, updated status text rendering for proper capitalization.

Added camelcase off to lint to accept the choice_type prop.

CEM-1026-Modify-KitchenCard-to-accept-choice_type_format


src/Containers/KitchenCard/KitchenCard.tsx
```diff
@@ -6,23 +6,21 @@ import { Status, StatusColors } from '../Status';
...
-export interface KitchenProps {
+export interface KitchenCardProps {
     customer: {
         name: string;
     };
     _id: string;
     items: Item[];
     orderType: string;
-    status: StatusColors;
+    statusColor: StatusColors;
```
ralph-dev, line 22: Having these two variables seperate could lead to showing the wrong color for a different status, wondering if there is a better way to make this error proof or if CDD requires this

src/Containers/KitchenCard/KitchenCard.tsx
```diff
@@ -61,7 +59,9 @@ export const KitchenCard: React.FC<KitchenProps> = ({
                 display="space-between"
                 padding="0 5px"
             >
-                <Status statusColor={status}>{status}</Status>
+                <Status statusColor={statusColor}>
+                    {status.charAt(0) + status.slice(1).toLowerCase()}
```
ralph-dev, line 63: Create a function for this, this will run every render, only depends on one prop

src/Containers/KitchenCard/KitchenCardItems.tsx
```diff
@@ -36,9 +31,18 @@ export const KitchenCardItems: React.FC<KitchenCardItemsProps> = ({
...
+                                (choice: Choice): React.ReactElement => (
+                                    <Paragraph margin="5px 0 0 20px">
+                                        {`${
+                                            choice.choice_type ===
+                                            ModifierChoiceTypeEnum.DEFAULT
+                                                ? ''
+                                                : choice.choice_type
+                                        } ${choice.name}`}
+                                    </Paragraph>
+                                ),
```
ralph-dev, line 45: Don't have inline functions inside your react tree, keep your react tree extremely small < 10 - 15 lines and use functions to do complex logic

ralph-dev (changes requested):
React Tree should be cleaner no inline methods / styling.

src/Containers/KitchenCard/KitchenCard.tsx
Ampatte2: Because the way the status component is using themeing to do "red" "green" "orange" an enum will need to be created to match the appropriate status color to the status being passed in.

I implemented this in the storybook. I think unless you error in typing the enum it will be error proof? I might be missing something.

StatusColorFilter

Ampatte2:
Unsure of the implementation of useCallback in kitchencarditems, please advise. Added additional enum to storybook to demonstrate status and statusColor error will be avoided.

src/Containers/KitchenCard/KitchenCard.tsx
ralph-dev: It's not ideal but it works i guess

src/Containers/KitchenCard/KitchenCard.tsx
```diff
@@ -48,6 +46,13 @@ export const KitchenCard: React.FC<KitchenProps> = ({
     TimeComponent,
     StatusModifierComponent,
 }): React.ReactElement => {
+    const statusFormat = useCallback(
+        (orderStatus: string) => {
```
ralph-dev, line 50: this doesn't need any paramteres becuase it can access status from function scope.

src/Containers/KitchenCard/KitchenCard.tsx
```diff
@@ -61,7 +66,9 @@ export const KitchenCard: React.FC<KitchenProps> = ({
                 display="space-between"
                 padding="0 5px"
             >
-                <Status statusColor={status}>{status}</Status>
+                <Status statusColor={statusColor}>
+                    {statusFormat(status)}
```
ralph-dev, line 70: no parameters

src/Containers/KitchenCard/KitchenCard.tsx
```diff
@@ -48,6 +46,13 @@ export const KitchenCard: React.FC<KitchenProps> = ({
     TimeComponent,
     StatusModifierComponent,
 }): React.ReactElement => {
+    const statusFormat = useCallback(
+        (orderStatus: string) => {
+            return orderStatus.charAt(0) + orderStatus.slice(1).toLowerCase();
+        },
+        [status],
```
ralph-dev, line 53: this is correct

ralph-dev (changes requested):
Small fix

ralph-dev (approved):
LGTM

PR #184, opened 2020-09-30
Chart Component

roggc:
Make a chart component with recharts library.

### DIFF
--- a/src/Containers/Chart.tsx
+++ b/src/Containers/Chart.tsx
@@ -159,13 +159,13 @@ const CustomTooltip: React.FC<CustomTooltipProps> = ({
     payload,
     ...props
 }): React.ReactElement => {
-    const formatValue = (value) => value.toFixed(2);
+    const getFormattedHoverValue = () => payload[0].value.toFixed(2);
 
     if (active) {
         return (
             <CustomTooltipBox active={active} payload={payload} {...props}>
                 <strong>
-                    $<Big>{`${formatValue(payload[0].value)}`}</Big>
+                    $<Big>{`${getFormattedHoverValue()}`}</Big>
                 </strong>
             </CustomTooltipBox>
         );
--- a/src/Containers/Chart.tsx
+++ b/src/Containers/Chart.tsx
@@ -1,12 +1,5 @@
 import React from 'react';
 import styled from 'styled-components';
-import { flex } from '../Utils/Mixins';
-import {
-    Main,
-    Responsive,
-    MainInterface,
-    ResponsiveInterface,
-} from '../Utils/BaseStyles';
 import {
     LineChart,
     Line,
@@ -15,6 +8,13 @@ import {
     Tooltip,
     CartesianGrid,
 } from 'recharts';
+import { flex } from '../Utils/Mixins';
+import {
+    Main,
+    Responsive,
+    MainInterface,
+    ResponsiveInterface,
+} from '../Utils/BaseStyles';
 
 export interface ChartProps extends ChartBoxProps {}
 
@@ -61,7 +61,7 @@ export const Chart: React.FC<ChartProps> = ({
             </Icons>
         </Header>
         <LineChart width={460} height={160} data={data}>
-            <XAxis dataKey={'label'} tickLine={false} />
+            <XAxis dataKey="label" tickLine={false} />
             <YAxis axisLine={false} tickLine={false} />
             <Tooltip
                 cursor={false}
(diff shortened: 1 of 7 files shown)


roggc:
chart

src/Containers/Chart.tsx
```diff
@@ -0,0 +1,270 @@
...
+const CustomTooltip: React.FC<CustomTooltipProps> = ({
+    active,
+    payload,
+    ...props
+}): React.ReactElement => {
+    if (active) {
+        return (
+            <CustomTooltipBox active={active} payload={payload} {...props}>
+                <strong>
+                    $<Big>{`${payload[0].value.toFixed(2)}`}</Big>
```
ralph-dev, line 166: Don't have inline functions in the react tree, the react tree should be very easy to read

ralph-dev (changes requested):
Make sure you react tree is clean, no inline functions or evaluations.
Also the design is missing some alignments but it's coming along really well.

src/Containers/Chart.tsx
roggc: I have tried to fix that in the next commit

src/Containers/Chart.tsx
```diff
@@ -159,11 +159,13 @@ const CustomTooltip: React.FC<CustomTooltipProps> = ({
...
     ...props
 }): React.ReactElement => {
+    const formatValue = (value) => value.toFixed(2);
+
     if (active) {
         return (
             <CustomTooltipBox active={active} payload={payload} {...props}>
                 <strong>
-                    $<Big>{`${payload[0].value.toFixed(2)}`}</Big>
+                    $<Big>{`${formatValue(payload[0].value)}`}</Big>
```
ralph-dev, line 168: you need to replace this entire line with {getFormattedHoverValue()}

roggc: Ok, thought would not work but it does. I fixed it in next commit

src/Containers/Chart.tsx
```diff
@@ -159,13 +159,13 @@ const CustomTooltip: React.FC<CustomTooltipProps> = ({
...
 }): React.ReactElement => {
-    const formatValue = (value) => value.toFixed(2);
+    const getFormattedHoverValue = () => payload[0].value.toFixed(2);
 
     if (active) {
         return (
             <CustomTooltipBox active={active} payload={payload} {...props}>
                 <strong>
-                    $<Big>{`${formatValue(payload[0].value)}`}</Big>
+                    $<Big>{`${getFormattedHoverValue()}`}</Big>
```
ralph-dev, line 168: Replace this whole line. in the React Tree i should only see
Either
1) <Component />
2) {GetComponent}
That way you can read the entire tree much clearer

roggc: I don't get it

ralph-dev: You need to make the tree readable

roggc: Yes, I understand this. I do not understand what changes I have to make to the code ...

roggc:
Screenshot 2020-10-01 at 15 31 18

src/Containers/Chart.tsx
```diff
@@ -0,0 +1,299 @@
...
+    ...props
+}): React.ReactElement => {
+    const getFormattedHoverValue = () => payload[0].value.toFixed(2);
+
+    if (active) {
+        return (
+            <CustomTooltipBox active={active} payload={payload} {...props}>
+                <strong>
+                    $
+                    <Big>{`${getFormattedHoverValue()}`}</Big>
```
ralph-dev, line 176:
```suggestion
                    <Big>${getFormattedHoverValue()}</Big>
```
Can't you just do this instead? If not, just cast the toFixed() with .toString()

ralph-dev (approved):
Ready for approval but code-wise the quality can use some work we will try another component.

src/Containers/Chart.tsx
roggc: I did that way because I needed the dollar sign be smaller than the number

src/Containers/Chart.tsx
```diff
@@ -0,0 +1,299 @@
...
+import { flex } from '../Utils/Mixins';
+import {
+    Main,
+    Responsive,
+    MainInterface,
+    ResponsiveInterface,
+} from '../Utils/BaseStyles';
+
+export interface ChartProps extends ChartBoxProps {}
+
```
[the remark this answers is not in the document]
roggc: ok...

PR #187, opened 2020-10-04
stock component

roggc:
stock

### DIFF
--- a/src/Containers/Stock.tsx
+++ b/src/Containers/Stock.tsx
@@ -1,0 +1,122 @@
+import React from 'react';
+import styled from 'styled-components';
+import { LineChart, Line } from 'recharts';
+import { ArrowUp } from '@styled-icons/entypo/ArrowUp';
+import { ArrowDown } from '@styled-icons/entypo/ArrowDown';
+import { flex } from '../Utils/Mixins';
+import {
+    Main,
+    Responsive,
+    MainInterface,
+    ResponsiveInterface,
+} from '../Utils/BaseStyles';
+import { MainTheme } from '../Themes';
+
+interface ChartDataItem {
+    value: string | number;
+}
+
+export interface StockProps extends StockBoxProps {
+    chartData: Array<ChartDataItem>;
+    chartColor: string;
+    title: string;
+    figure: number;
+    rate: number;
+    bgColor: string;
+}
+
+const getChartColor = (colorKey: string) =>
+    MainTheme.colors[colorKey] || colorKey;
+
+const getFigure = (figure: number) =>
+    figure.toLocaleString(undefined, { maximumFractionDigits: 2 });
+
+const getRateIcon = (rate: number) =>
+    rate < 0 ? <StyledArrowDown /> : <StyledArrowUp />;
+
+export const Stock: React.FC<StockProps> = ({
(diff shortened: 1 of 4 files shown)


src/Containers/Stock.tsx
```diff
@@ -0,0 +1,122 @@
...
+    bgColor: string;
+}
+
+const getChartColor = (colorKey: string) =>
+    MainTheme.colors[colorKey] || colorKey;
+
+const getFigure = (figure: number) =>
+    figure.toLocaleString(undefined, { maximumFractionDigits: 2 });
+
+const getRateIcon = (rate: number) =>
```
ralph-dev, line 34: Shouldn't this be inside the component?

ralph-dev (commented):
Good job catching the array of values! Small question about the functions and why there not in the component?

src/Containers/Stock.tsx
roggc: well, they have no dependencies on props so they can be outside, so I thought why not place them outside, in that way in each render there is no redefinition of the function, but I don't really know where it is better to place them.

Ampatte2: It is being implemented in the render so it will run each time. Each of the args for the functions are in the list of props. UseCallback would be good here to not run the function on each render if props are the same. I would change getChartColor arg colorKey to chartColor for readability.

src/Containers/Stock.tsx
```diff
@@ -0,0 +1,122 @@
...
+const Content = styled.div`
+    margin: 10px 20px 10px 10px;
+`;
+
+const StyledArrowUp = styled(ArrowUp)`
+    width: 15px;
+`;
+const StyledArrowDown = styled(ArrowDown)`
+    width: 15px;
+`;
```
Ampatte2, line 120: Don't Repeat Yourself. This can be condensed to styled.svg` width: 15px` then in the implementation StyledIcon as={icon you want}

src/Containers/Stock.tsx
```diff
@@ -0,0 +1,122 @@
...
+interface TextProps extends React.HTMLAttributes<HTMLDivElement> {
+    size: string;
+}
+
+const Text = styled.div<TextProps>`
+    ${({ theme, ...props }): string => `
+font-size:${theme.font.size[props.size] || props.size};
+color:${theme.colors[props.color as string] || props.color};
+`}
+`;
```
Ampatte2, line 109: There are three different Text elements in the React-UI library: Paragraph, SmallText, and Heading. No need to create a text element for each component you make. Use other components from the library when you can.

Ampatte2 (changes requested):

src/Containers/Stock.tsx
roggc: I know. But these you mention are all p tags in the end, and a p tag has a lot of margins vertically, and I needed less margin, so I implemented my own version for this reason. I remember you already commented on this point to me in the previous PR.

Ampatte2: With the type attribute the components can be typed as div.
Type change
Screenshot from 2020-10-05 16-24-32

Ampatte2 (approved):
LGTM

PR #188, opened 2020-10-09
CEM-1054 Add Functionality To List

Ampatte2:
KitchenCard needed stronger typing for the data it was recieving. Added constants to KitchenCard.

List properties have been renamed for clarity, and ListToggle has been made it's own component. 

In ListHeader the option to add a component to the headerRow instead of an Icon has been added.

ListItem has a isSelected prop to indicate that an item is currently selected.

Navigation has an added media query to help with consistent styling of a collapsable list at tablet media query.

CEM-1054-Add-Functionality-To-List

### DIFF
--- a/src/Containers/KitchenCard/constants.ts
+++ b/src/Containers/KitchenCard/constants.ts
@@ -1,19 +1,17 @@
-export interface Choice{
+export interface ModifierChoiceInput {
     name: string;
     choice_type: string;
 }
-
-export interface ItemModifier {
+export interface OrderModifier {
     name?: string;
-    choices: Choice[];
+    choices: ModifierChoiceInput[];
 }
-
-export interface Item {
+export interface OrderItem {
     name: string;
-    modifiers: ItemModifier[];
+    modifiers: OrderModifier[];
 }
-export enum ModifierChoiceTypeEnum {
-    ADD = 'ADD',
-    NO = 'NO',
-    DEFAULT = 'DEFAULT'
+export declare enum ModifierChoiceTypeEnum {
+    ADD = "ADD",
+    NO = "NO",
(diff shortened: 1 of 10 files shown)


src/Containers/KitchenCard/constants.ts
```diff
@@ -1,19 +1,17 @@
...
+export interface OrderItem {
     name: string;
-    modifiers: ItemModifier[];
+    modifiers: OrderModifier[];
 }
-export enum ModifierChoiceTypeEnum {
-    ADD = 'ADD',
-    NO = 'NO',
-    DEFAULT = 'DEFAULT'
+export declare enum ModifierChoiceTypeEnum {
```
ralph-dev, line 13: Can't you import from TS SDK? @DenisMela can we move this

src/Containers/List/List.tsx
```diff
@@ -28,24 +28,25 @@ export const List: React.FC<ListProps> = ({
...
-                    setIsToggled(true);
-                    break;
-                case 'swipeLeft':
-                    setIsToggled(false);
-                    break;
-                default:
-                    break;
+            if (setIsOpen) {
+                switch (type) {
+                    case 'swipeRight':
```
ralph-dev, line 41: Labelling this or using the enums from the lib is stronger but this works

ralph-dev (approved):
Lgtm

src/Containers/KitchenCard/constants.ts
ralph-dev: I mean it shouldn't becuase this shouldn't touch data but maybe we should separate TS SDK with an @ types lol I learned something new today. This is ok then

PR #194, opened 2020-10-15
Converted header row story from .js to .tsx (v6)

williancozzi:
Got the props from the HeaderRow.tsx and turned HeaderRow.stories.js into HeaderRow.stories.tsx

image


ralph-dev (approved):
Next time don't delete the file just rename for proper GitHub diff to occur. Try another file of your choice in the next PR

PR #196, opened 2020-10-16
converted file from js to tsx and make the changes

williancozzi:
Renamed the file from .js to .tsx;
made the current changes in the component into v6

image


ralph-dev (approved):
LGTM! Try another component

PR #197, opened 2020-10-16
CEM-1057_Featured_Profile_Card

cavingayle:
Created Featured Profiles Card component which can be seen in the image below.

Screen Shot 2020-10-16 at 10 17 38 AM


src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -0,0 +1,125 @@
...
+        </Container>
+    );
+};
+
+interface CircleImageProps {
+    background: string;
+}
+
+const CircleImage = styled.li<CircleImageProps>`
+    display: flex;
```
ralph-dev, line 79: Mixins?

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -0,0 +1,125 @@
...
+                .map(
+                    (profile, index): React.ReactElement =>
+                        getProfileCircles(index, profile),
+                )
+                .splice(0, 7)}
+        </Container>
+    );
+};
+
+interface CircleImageProps {
```
ralph-dev, line 74: interface ICircleImage {

}

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -0,0 +1,125 @@
...
+        return (
+            <CircleImage background="grey">
+                <Icon as={AddUser} />
+            </CircleImage>
+        );
+    };
+
+    return (
+        <Container>
+            {profileData
```
ralph-dev, line 64: no inline methods in the tree

ralph-dev: replace with {getProfiles()}

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -0,0 +1,125 @@
...
+                .map(
+                    (profile, index): React.ReactElement =>
+                        getProfileCircles(index, profile),
+                )
+                .splice(0, 7)}
```
ralph-dev, line 69: 7 is a magic number, what does this mean?

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -0,0 +1,125 @@
...
+            );
+        }
+        if (index > 2 && index < 5) {
+            return (
+                <CircleImage background="orange">
+                    <CircleContent>{profile.initials}</CircleContent>
+                </CircleImage>
+            );
+        }
+        if (index > 4 && index < 6) {
```
ralph-dev, line 48: Magic numbers

ralph-dev (changes requested):
Code quality needs improvement:

- Start interfaces with I
- No Magic Numbers, put const VIEWABLE_PROFILE_LIMIT = 3 etc;

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -3,80 +3,89 @@ import styled from 'styled-components';
 import { AddUser } from '@styled-icons/entypo/AddUser';
 import { flex } from '../Utils/Mixins';
 
-export interface profile {
+export interface Iprofile {
```
ralph-dev, line 6: IProfile

P should be capital it's a new word

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -3,80 +3,89 @@ import styled from 'styled-components';
...
+                        index > PROFILE_INITIALS_START_INDEX &&
+                        index < PROFILE_INITIALS_END_INDEX
+                    ) {
+                        return (
+                            <CircleImage background="orange">
+                                <CircleContent>
+                                    {profile.initials}
+                                </CircleContent>
+                            </CircleImage>
+                        );
```
ralph-dev, line 59: Does this look like a readable code?
No. Refactor it into smaller functions so you can understand what's going on.

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -3,80 +3,89 @@ import styled from 'styled-components';
...
+        const REMAINING_PROFILES_VALUE = profileData.length - 5;
+        const MAX_PROFILES = 7;
+        const START_FROM_PROFILE_INDEX = 0;
+
+        return profileData
+            .map(
+                (profile, index): React.ReactElement => {
+                    if (index < PROFILE_PICTURE_LIMIT) {
+                        return (
+                            <CircleImage background="none" key={profile.image}>
```
ralph-dev, line 39: The word circleImage is repeated multiple times, this is not DRY (Do Not Repeat Yourself)

Maybe use CircleStyle and do {...CircleStyle} so you only have to define once? Something to look into :)

ralph-dev (changes requested):
Codes looking better, still a bit more to go!

src/Containers/FeaturedProfilesCard.tsx
cavingayle: I'd love to have some more feedback on this point. You're definitely right about the code not being dry. Since CircleImage is just a list item that I'm returning conditionally based on the current index I'm not sure how to make it drier. I'm wondering if it's the way I decided to structure the code that's limiting how dry I can make it? What would be the best approach to solving this?

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -0,0 +1,146 @@
...
+}
+
+export const FeaturedProfilesCard: React.FC<IFeaturedProfilesCardProps> = ({
+    alt,
+    profileData,
+    width = 100,
+    height = 100,
+}): React.ReactElement => {
+    const getProfileCircles = (): any => {
+        const PROFILE_PICTURE_LIMIT = 3;
```
ralph-dev, line 26: this has nothing to do with the componet it should be at top of the file

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -0,0 +1,146 @@
...
+        const PROFILE_INITIALS_START_INDEX = 2;
+        const PROFILE_INITIALS_END_INDEX = 5;
+        const REMAINING_PROFILES_INDEX = 5;
+        const REMAINING_PROFILES_VALUE = profileData.length - 5;
```
ralph-dev, line 30: This can stay inside though, just lowercase since it's not a TRUE constant, it's a dynamic variable so lowercase all

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -0,0 +1,146 @@
...
+            .map(
+                (profile, index): React.ReactElement => {
+                    if (index < PROFILE_PICTURE_LIMIT) {
+                        return UserProfilePicture(profile);
+                    }
+                    if (
+                        index > PROFILE_INITIALS_START_INDEX &&
+                        index < PROFILE_INITIALS_END_INDEX
+                    ) {
+                        return UserInitials(profile);
```
ralph-dev, line 81: why is the function in caps?
Please look at the JavaScript Style Guide on Confluence

ralph-dev (changes requested):
Code Quality

src/Containers/FeaturedProfilesCard.tsx
ralph-dev: The dynamic variables are the props, thus create a props object and spread it one the CircleImage since that is never changing.

src/Containers/FeaturedProfilesCard.tsx
cavingayle: Does this mean the getProfilesCircles function does not need to be in the FeaturedProfilesCard component? I currently have getProfileCircles accessing props from the component.

ralph-dev: I mean the variable 3 does not need to be inside the function

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -0,0 +1,147 @@
...
+    height?: number;
+}
+
+export const FeaturedProfilesCard: React.FC<IFeaturedProfilesCardProps> = ({
+    alt,
+    profileData,
+    width = 100,
+    height = 100,
+}): React.ReactElement => {
+    const getProfileCircles = (): any => {
```
ralph-dev, line 32: any is not good, is this ReactElement or ReactNode?

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -0,0 +1,147 @@
...
+        const userProfilePicture = (profile: any): React.ReactElement => {
```
ralph-dev, line 33: any is not good, is this object or string etc?

ralph-dev (commented):
@Ampatte2 Can review / approve this PR I am not approving it until it is more clear.

cavingayle:
> @Ampatte2 Can review / approve this PR I am not approving it until it is more clear.

@Ampatte2 yes can you let me know all the things wrong before I push again so I can fix them.

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -0,0 +1,147 @@
...
+                    ) {
+                        return userInitials(profile);
+                    }
+                    if (index === REMAINING_PROFILES_INDEX) {
+                        return remainingUserProfiles();
+                    }
+                    return addUserIcon();
+                },
+            )
+            .splice(START_FROM_PROFILE_INDEX, MAX_PROFILES);
```
Ampatte2, line 90: This logic here needs some fixes. 

Creating a excess addUserIcons to splice them off is not good. Would like to see additional logic for handling the length of for the remainingUserProfiles function, and then splicing just what you need before mapping.

The case here which isn't handled is what if I have 5 or less profiles?

src/Containers/FeaturedProfilesCard.tsx
Ampatte2, line 32: Using useCallback for this function would be sweet.

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -0,0 +1,147 @@
...
+            );
+        };
+
+        const addUserIcon = (): React.ReactElement => {
+            return (
+                <CircleImage background="grey">
+                    <Icon as={AddUser} />
+                </CircleImage>
+            );
+        };
```
Ampatte2, line 70: Is this supposed to be functional?

Ampatte2 (changes requested):
This works and looks good, but is not extensible. There could be a case that I want more or less user profiles displayed. And if I have less profiles, like on a new account, I lose the add feature.

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -21,18 +21,27 @@ export interface IFeaturedProfilesCardProps {
     alt: string;
     width?: number;
     height?: number;
+    onClick?: Function;
```
ralph-dev, line 24: Function should never be used 
try
() => void
with proper typing ^

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -21,18 +21,27 @@ export interface IFeaturedProfilesCardProps {
...
 }
 
 export const FeaturedProfilesCard: React.FC<IFeaturedProfilesCardProps> = ({
+    onClick = (): void => {
```
ralph-dev, line 28: This should be typed by typescript interface above not inline

ralph-dev:
Resolve conflicts from master btw ^

src/Containers/FeaturedProfilesCard.tsx
Ampatte2: bump.

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -21,140 +22,118 @@ export interface IFeaturedProfilesCardProps {
...
+                            <CircleContent>{profile.initials}</CircleContent>
+                        </FeaturedProfile>
+                    );
+                default:
+                    return (
+                        <FeaturedProfile profile={profile} background="gray">
+                            <CircleContent>
+                                {profileData.length - profileInitialsEndIndex}
+                            </CircleContent>
+                        </FeaturedProfile>
```
Ampatte2, line 84: FeatureProfile is repeated here three times. You are returning either an image or a circle content, FeatureProfile can handle this through props. In the end this will look like <FeaturedProfile whateverprops />. instead of this whole thing repeated three times.

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -21,140 +22,118 @@ export interface IFeaturedProfilesCardProps {
...
-    onClick = (): void => {
-        return undefined;
-    },
     alt,
     profileData,
     width = 100,
     height = 100,
 }): React.ReactElement => {
     const getProfileCircles = useCallback(() => {
+        const getProfilePictureLimit = () => {
```
Ampatte2, line 34: Most of the naming here is great, little nitpick. Get is usually associated with making an request or fetching some data. Maybe renderProfileCircles is more appropriate.

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -21,140 +22,118 @@ export interface IFeaturedProfilesCardProps {
...
 export const FeaturedProfilesCard: React.FC<IFeaturedProfilesCardProps> = ({
-    onClick = (): void => {
-        return undefined;
-    },
     alt,
     profileData,
     width = 100,
     height = 100,
 }): React.ReactElement => {
     const getProfileCircles = useCallback(() => {
```
Ampatte2, line 33: Most of the naming here is great, little nitpick. Get is usually associated with making an request or fetching some data. Maybe renderProfileCircles is more appropriate.

src/Containers/FeaturedProfilesCard.tsx
Ampatte2, line 34: I like line 41 determineMaxProfileLength, let's do that again here.

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -21,140 +22,118 @@ export interface IFeaturedProfilesCardProps {
...
-                <CircleImage background="orange" key={profile.id}>
-                    <CircleContent>{profile.initials}</CircleContent>
-                </CircleImage>
-            );
-        };
+        const profileInitialsEndIndex =
+            profiles.length - PROFILE_INITIALS_INDEX_ENDPOINT;
 
-        const remainingUserProfiles = (
+        const getProfiles = (
```
Ampatte2, line 56: Same as above, but for this one it needs to be specific to what it's doing.

stories/Containers/FeaturedProfilesCard.stories.tsx
```diff
@@ -73,6 +73,7 @@ export default {
     args: {
         profileData: profiles,
         alt: 'Profile Image',
+        maxProfilesDisplay: 6,
```
Ampatte2, line 76: Leftover from testing?

src/Containers/FeaturedProfile.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const FeaturedProfile: React.FC<IFeaturedProfileProps> = ({
+    children,
+    background,
+    profile,
+}) => {
+    return (
+        <CircleImage background={background} key={profile.id}>
+            {children}
+        </CircleImage>
+    );
```
Ampatte2, line 29: I don't see the need for two seperate components here, they perform the same function. Props can determine what to display.

Ampatte2 (changes requested):
I like this, there are a few changes that can be made to improve code quality/readibility. There was one issue with the component itself: Testing different props for width and height this surfaced.

issues with width and height

src/Containers/FeaturedProfile.tsx
```diff
@@ -39,9 +43,22 @@ const CircleImage = styled.li<ICircleImageProps>`
     align-items: center;
     border: solid white 3px;
     border-radius: 50%;
-    object-fit: cover;
     overflow: auto;
     width: 100px;
     height: 100px;
```
Ampatte2: Should these take the same props as image?

src/Containers/FeaturedProfilesCard.tsx
```diff
@@ -58,30 +54,30 @@ export const FeaturedProfilesCard: React.FC<IFeaturedProfilesCardProps> = ({
...
                 default:
                     return (
-                        <FeaturedProfile profile={profile} background="gray">
-                            <CircleContent>
-                                {profileData.length - profileInitialsEndIndex}
-                            </CircleContent>
-                        </FeaturedProfile>
+                        <FeaturedProfile
+                            key={profile.id}
+                            profilesRemaining={profileInitialsEndIndex}
```
Ampatte2, line 75: this should be the remaining profiles.

Ampatte2 (changes requested):
Two changes, but after that LGTM.

Ampatte2 (approved):
LGTM

PR #198, opened 2020-10-18
converted input stories from js to tsx

williancozzi:
Made the changes to turn input.stories.js into .tsx

image


ralph-dev (approved):
LGTM

PR #203, opened 2020-10-21
Added React.Element to LabelLayoutProps Description

dannybuii:
LabelLayoutProps description takes in a string originally. However, some components being used in the Cheapreats Dashboard are actually React.Element. Therefore React.Element needed to be added as a type in the description.

### DIFF
--- a/src/Fragments/LabelLayout.tsx
+++ b/src/Fragments/LabelLayout.tsx
@@ -21,7 +21,7 @@ export interface LabelLayoutProps
         ImplicitPropsInterface {
     name?: string;
     label?: string;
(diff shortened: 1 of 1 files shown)


ralph-dev (approved):
LGTM, is it react Element or React Node btw?

ralph-dev:
Becuase I think you can replace it all with react node?

dannybuii:
It's an react element. Because before I added it in, Typescript was throwing error saying "react element can't be string"

ralph-dev:
@dannybuii yes but strin | Element 
Means you can just do react node

PR #211, opened 2020-10-29
CEM-1176 Add Validation to MaskedInput

Ampatte2:
MaskedInput had an issue that min and max weren't being implemented on the component, and errors were not given for being outside of the min max.

MaskedInput now has validation and error messaged based on if the value is LT or GT the min and max values respectively.

MaskedInput


stories/Containers/List.stories.tsx
```diff
@@ -57,8 +65,8 @@ export default {
 } as Meta;
 
 export const Basic: Story<ListProps> = (args) => {
-    const [isOpen, setIsOpen] = useState(true);
-
+    const [{ isOpen }, updateArgs] = useArgs();
```
ralph-dev, line 68: I like this a lot.

ralph-dev (approved):
LGTM, just note that 100 is for percentage and 100 00 is value becuase of the /100. I don't think anything needs to change just want to highlight this for Jun

roggc (commented):
lgtm.

stories/Containers/List.stories.tsx
Ampatte2: There was a solution!

ralph-dev: @williancozzi He found the solution

PR #212, opened 2020-10-31
VRBWA-132 Middle Canvas

xsharonhe:
Changing pointer events for the div:
Middle Canvas

Adding responsiveness for small phone:
Middle Canvas Phone

Added rearranging component of the list and labels within the list:
Middle Canvas


ralph-dev:
@xsharonhe let us jump on a call to explain som concepts tomorrow

src/Containers/MiddleCanvas/DroppableContainerContents.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import styled from 'styled-components';
+import { MainInterface, ResponsiveInterface } from '../../Utils/BaseStyles';
+import { Mixins } from '../../Utils';
+
+export interface DroppableContainerContentsProps extends MainInterface, ResponsiveInterface, React.HTMLAttributes<HTMLDivElement> {
+    styling: string,
```
ralph-dev, line 7: remove this entirely, and just style through te props spread

src/Containers/MiddleCanvas/DroppableContainerContents.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export const DroppableContainerContents: React.FC<DroppableContainerContentsProps> = ({
+    styling,
+    droppableLabel,
+    ...props
+}): React.ReactElement => {
+    return (
+        <Wrapper display={styling} {...props}>
+            {Object.values(droppableLabel).map(label => (
```
ralph-dev, line 24: why object.values if string array

src/Containers/MiddleCanvas/DroppableContainerContents.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { MainInterface, ResponsiveInterface } from '../../Utils/BaseStyles';
+import { Mixins } from '../../Utils';
+
+export interface DroppableContainerContentsProps extends MainInterface, ResponsiveInterface, React.HTMLAttributes<HTMLDivElement> {
+    styling: string,
+    droppableLabel: string[],
+    isColumn: boolean
+};
+
+interface WrapperProps {
```
ralph-dev, line 12: Co-Location

src/Containers/MiddleCanvas/DroppableContainerContents.tsx
```diff
@@ -0,0 +1,42 @@
...
+                    {label}
+                </div>
+            ))}
+        </Wrapper>
+    );
+}
+
+const Wrapper = styled.div<WrapperProps>`
+    display: flex;
+    ${(props): string | undefined =>
```
ralph-dev, line 35: (props) -> ( { display } )

src/Containers/MiddleCanvas/DroppableContainerContents.tsx
```diff
@@ -0,0 +1,42 @@
...
+        props.display && Mixins.flex(props.display)};
+    ${({ isColumn }): string => `
+        ${isColumn ? Mixins.flex('column') : ''};
+        padding: ${isColumn ? '5px' : ''};
```
ralph-dev, line 39: abstract to a css``;
isColumn : true -> css``;
false -> undefined

https://github.com/cheapreats/mobile/blob/master/src/components/atoms/Circle.tsx

src/Containers/MiddleCanvas/DroppableContainerContents.tsx
```diff
@@ -0,0 +1,42 @@
...
+    `};
+    padding: 0 10px;
```
ralph-dev, line 41: watch out, padding overrirdes

src/Containers/MiddleCanvas/DroppableElement.tsx
```diff
@@ -0,0 +1,78 @@
...
+    Droppable, 
+    DroppableProvided, 
+    DroppableStateSnapshot 
+} from 'react-beautiful-dnd';
+import { StyledIcon } from 'styled-icons/types';
+import { DroppableContainerContents } from './DroppableContainerContents';
+import { MainInterface, ResponsiveInterface } from '../../Utils/BaseStyles';
+
+export interface DroppableElementProps extends MainInterface, ResponsiveInterface, React.HTMLAttributes<HTMLDivElement> {
+    droppableLabels: string[][] | StyledIcon [][],
```
ralph-dev, line 13:
```suggestion
    droppableLabels: string[][] | StyledIcon[][],
```

src/Containers/MiddleCanvas/DroppableElement.tsx
```diff
@@ -0,0 +1,78 @@
...
+    styling: string,
```
ralph-dev, line 14: dont do this pattern

src/Containers/MiddleCanvas/DroppableElement.tsx
```diff
@@ -0,0 +1,78 @@
...
+    isDragging?: boolean,
+    isColumn?: boolean,
+    isPreview?: boolean
+};
+
+const FIRST_LABEL = 0;
+const DESIGN_BACKGROUND_COLOR = '#ffffff';
+const PREVIEW_BACKGROUND_COLOR = '#f5f5f5';
+const DESIGN_COLOR = '#b7b7b7';
+const PREVIEW_COLOR = 'black';
```
ralph-dev, line 29: Does not work with Themeing , what is the Light / Dark Mode

src/Containers/MiddleCanvas/DroppableElement.tsx
```diff
@@ -0,0 +1,78 @@
...
+export const DroppableElement: React.FC<DroppableElementProps> = ({
+    droppableLabels,
+    styling,
+    isColumn,
+    isPreview,
+    ...props
+}): React.ReactElement => {
+    return (
+        <Wrapper {...props}>
+            {Object.values(droppableLabels).map(droppableLabel => (
```
ralph-dev, line 40: no need

ralph-dev: Try to avoid writing logic instead the tree.

<Wrapper>
{getDraggableObject()}
</Wrapper>

src/Containers/MiddleCanvas/DroppableElement.tsx
```diff
@@ -0,0 +1,78 @@
...
+    width: 349px;
+    border-radius: 4px;
+    ${({ isColumn }): string => `
+        padding: ${isColumn ? '10px 0 2px 0' : '10px 0 0 0'};
+    `};
+    margin: 10px;
+    ${({ isPreview }): string => `
+        background-color: ${isPreview ? DESIGN_BACKGROUND_COLOR : PREVIEW_BACKGROUND_COLOR};
+        color: ${isPreview ? DESIGN_COLOR : PREVIEW_COLOR}
+        padding: ${isPreview ? '0px 0 0 0' : '10px 0 0 0'};
```
ralph-dev, line 75: no px for 0

src/Containers/MiddleCanvas/DualSelectBar.tsx
```diff
@@ -0,0 +1,66 @@
...
+    selectedOption,
+    setSelectedOption,
+    ...props
+}): React.ReactElement => {
+    return (
+        <Wrapper {...props}>
+            <Option 
+                isSelected={selectedOption === leftSelectOption}
+                onClick={() => setSelectedOption(leftSelectOption)}
+            >
```
ralph-dev, line 32: Abstract the duplciate logic with either an object spread
Function that returns an object spread

const getOptionFields(Left) getOptionFields(Right) => {
 return {
         isSelected={selectedOption === leftSelectOption}
                onClick={() => setSelectedOption(leftSelectOption)}
}
}

src/Containers/MiddleCanvas/DualSelectBar.tsx
```diff
@@ -0,0 +1,66 @@
...
+                { leftSelectOption }
+            </Option>
+            <Option> | </Option>
```
ralph-dev, line 35: Option or View?

ralph-dev: <View  />

width: 1px
border-color: black

ralph-dev: https://www.w3schools.com/howto/howto_css_vertical_line.asp

src/Containers/MiddleCanvas/DualSelectBar.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+const Wrapper = styled.div`
+    font-size: 12px;
+    font-weight: bold;
+    font-stretch: normal;
+    font-style: normal;
+    line-height: 1.25;
+    letter-spacing: normal;
+    display: flex;
```
ralph-dev, line 53: mixins

src/Containers/MiddleCanvas/DualSelectRadio.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+interface TextProps {
+    isRightOption?: boolean;
+};
+
+interface SelectContainerProps {
+    isVisible?: boolean;
+};
+
+export interface dualSelectOption {
```
ralph-dev, line 32: don't reference before initiliaztion

src/Containers/MiddleCanvas/DualSelectRadio.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+export const DualSelectRadio: React.FC<DualSelectRadioProps> = ({
+    caption = 'Dual Select',
+    leftPlaceholder = 'Light/Sound',
+    headerSpacingStyle = 'space-between',
+    dualSelectOptions,
+    ...props
+}): React.ReactElement => {
+    const firstSelectOption = Object.values(dualSelectOptions)[FIRST_SELECT_OPTION].title;
+    const [leftSelectOption, setLeftSelectOption] = useState('Flash Once');
```
ralph-dev, line 53: Fixed Number Types, Use an ENUM.

src/Containers/MiddleCanvas/DualSelectRadio.tsx
```diff
@@ -0,0 +1,150 @@
...
+            <Header>
+                {caption}
+            </Header>
+            <SelectArea>
+                <Row display={headerSpacingStyle}>
+                    <Text>
+                        {leftPlaceholder}
+                    </Text>
+                    <Container>
+                        <Text isRightOption>
```
ralph-dev, line 76: ={true}

src/Containers/MiddleCanvas/MiddleCanvasElements.ts
```diff
@@ -0,0 +1,15 @@
+export interface ITemplatePrefill {
+    [key: string]: {
+        title: string,
+        labels: string[][],
+        display: string,
+        column: boolean
+    };
+};
+
+export interface IPrinterOptions {
```
ralph-dev, line 10: See if Type Duplication Code can be abstracted into a more general type

src/Containers/MiddleCanvas/Template.tsx
```diff
@@ -0,0 +1,86 @@
...
+        return templatePrefill.labels;
+    });
+
+    const templatePrefillDisplays = Object.values(templatePrefills).map((templatePrefill: ITemplatePrefill) => {
+        return templatePrefill.display;
+    });
+
+    const templatePrefillColumn = Object.values(templatePrefills).map((templatePrefill: ITemplatePrefill) => {
+        return templatePrefill.column;
+    });
```
ralph-dev, line 41: O(4N)

src/Containers/MiddleCanvas/Template.tsx
```diff
@@ -0,0 +1,86 @@
...
+                    <>
+                        {!isPreview && (
+                            <Header>
+                                {templatePrefillTitles[index]}
+                            </Header>
+                        )}
+                        <DroppableElement
+                            droppableLabels={templatePrefillLabels[index]}
+                            styling={templatePrefillDisplays[index]}
+                            isColumn={templatePrefillColumn[index]}
```
ralph-dev, line 59: templatePrefill.column? remove maps at the top

ralph-dev (changes requested):
Code quality

ralph-dev:
@xsharonhe Once this is fixed, let's tag Andrew to review it

src/Containers/MiddleCanvas/DroppableContainerContents.tsx
xsharonhe: This is the comment I meant. How would I know when to change it from space-between to centre (which is what I need for the different displays) without an indicator to do so?

src/Containers/MiddleCanvas/DualSelectBar.tsx
```diff
@@ -0,0 +1,71 @@
...
+                onClick={getOptionFields(rightSelectOption).onClick}
+            >
+                { rightSelectOption }
+            </Option>
+        </Wrapper>
+    );
+}
+
+const Wrapper = styled.div`
+    font-size: 12px;
```
ralph-dev, line 47: Should be coming from the theme Object for consistency

src/Containers/MiddleCanvas/DualSelectBar.tsx
```diff
@@ -0,0 +1,71 @@
...
+    margin: 5px;
+    padding: 5px;
+    ${({ theme, isSelected }): string => `
+        color: ${isSelected ? theme.colors.primary : theme.colors.border};
+        border-bottom: ${isSelected ? `solid 1px ${theme.colors.primary}` : ''};
+    `}
+`;
+
+const Divider = styled.div`
+    border-left: 1px solid black;
```
ralph-dev, line 69: Colour should be coming from theme Object

src/Containers/MiddleCanvas/DualSelectRadio.tsx
```diff
@@ -0,0 +1,138 @@
...
+                        /> 
+                    ))}
+                </SelectContainer>
+            </SelectArea>
+        </Wrapper>
+    );
+}
+
+const Wrapper = styled.div`
+    font-size: 12px;
```
ralph-dev, line 73: Theme

src/Containers/MiddleCanvas/TableRow.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const TableRow: React.FC<TableRowProps> = ({
+    labels,
+    ...props
+}): React.ReactElement => {
+    return (
+        <Wrapper {...props}>
+            <Tr>
+                <td>
+                    {Object.values(labels).map(label => (
```
ralph-dev, line 17: Do you need object.values? Can't you just map the labels.

ralph-dev (commented):
Stop using object values and just map things :)

ralph-dev:
Note this screen has a few use cases, I think you are missing one.

1) Ability to re-order entire section groups at once (Check)

2) Ability to re-order elements inside a section (Uncheck?)

src/Containers/MiddleCanvas/Template.tsx
```diff
@@ -0,0 +1,159 @@
...
+     * @param {ITemplatePrefill} list - list of objects to reorder
+     * @param {number} startIndex - index of where the element originates from
+     * @param {number} endIndex - index of where the element will be placed
+     */
+    const reorder = (list: ITemplatePrefill, startIndex: number, endIndex: number): ITemplatePrefill => {
+        const result = Object.values(list);
+        const [removed] = result.splice(startIndex, 1);
+        result.splice(endIndex, 0, removed);
+    
+        return result;
```
Ampatte2, line 40: Naming can be better here, result does not tell me much about what is happening.

Also 1, 0 are magic numbers.

src/Containers/MiddleCanvas/DualSelectBar.tsx
```diff
@@ -0,0 +1,71 @@
...
+    selectedOption,
+    setSelectedOption,
+    ...props
+}): React.ReactElement => {
+    const getOptionFields = (field: string) => {
+        return ({
+            isSelected: selectedOption === field,
+            onClick: () => setSelectedOption(field)
+        });
+    };
```
Ampatte2, line 25: This could be wrapped in a usecallback.

src/Containers/MiddleCanvas/DualSelectRadio.tsx
```diff
@@ -0,0 +1,138 @@
...
+    caption = 'Dual Select',
+    leftPlaceholder = 'Light/Sound',
+    headerSpacingStyle = 'space-between',
+    dualSelectOptions,
+    ...props
+}): React.ReactElement => {
+    const firstSelectOption = Object.values(dualSelectOptions)[FIRST_SELECT_OPTION].title;
+    const [leftSelectOption, setLeftSelectOption] = useState(Object.values(dualSelectOptions)[FIRST_SELECT_OPTION].labels[FIRST_LABEL]);
+    const [rightSelectOption, setRightSelectOption] = useState(Object.values(dualSelectOptions)[SECOND_SELECT_OPTION].labels[FIRST_LABEL]);
+    const [isCollapsed, setIsCollapsed] = useState(false);
```
Ampatte2, line 31: Object.values three times on the same prop doesn't seem right. Is there a better way to initialize this component?

src/Containers/MiddleCanvas/DualSelectRadio.tsx
```diff
@@ -0,0 +1,138 @@
...
+                    </Container>
+                </Row>
+                <SelectContainer isVisible={isCollapsed}>
+                    {!!isCollapsed && Object.values(dualSelectOptions).map((printOption): React.ReactElement => (
+                        <RadioOptions
+                            title={printOption.title}
+                            labels={printOption.labels} 
+                            firstSelectOption={firstSelectOption}
+                            setLeftSelectOption={setLeftSelectOption}
+                            setRightSelectOption={setRightSelectOption}
```
Ampatte2, line 63: This gives me TS errors on my branch.

src/Containers/MiddleCanvas/Template.tsx
```diff
@@ -0,0 +1,159 @@
...
+import { 
+    DragDropContext, 
+    Droppable,
+    DroppableProvided,
+    Draggable,
+    DraggableProvided,
+    DraggableStateSnapshot,
+    DropResult,
+} from 'react-beautiful-dnd';
+import { ITemplatePrefill } from '/MiddleCanvasTypes';
```
Ampatte2, line 12: This should be './MiddleCanvasTypes' when changed I get a lot of 6 TS errors in this file.

Ampatte2 (commented):
You have made some nice radio components dual select and such. These are components in their own right and should be usable through the CE react ui library. I would export/make stories for them and put them in appropriate folders.

For the UI on both dual select radio dropdowns you could maybe use transitions to make the dropdown flow smoother. Look into the Select component/labellayout for how they implemented and use that.

Am i supposed to be able to interact with the upload/insert fields?

This dropdown only being half the div is odd to me, what do you think @ralph-dev ?

dropdown

src/Containers/MiddleCanvas/Template.tsx
xsharonhe: Ah, it is because I made a general type string[] | string[][] are not really the same. I will remove the general type and split into two types. Thanks

xsharonhe:
> You have made some nice radio components dual select and such. These are components in their own right and should be usable through the CE react ui library. I would export/make stories for them and put them in appropriate folders.
> 
> For the UI on both dual select radio dropdowns you could maybe use transitions to make the dropdown flow smoother. Look into the Select component/labellayout for how they implemented and use that.
> 
> Am i supposed to be able to interact with the upload/insert fields?
> 
> This dropdown only being half the div is odd to me, what do you think @ralph-dev ?
> 
> dropdown

Will do thanks for that advice.

The middle canvas is supposed to interact with other components, as in you should be able to drag things into these spaces, but the "upload image" is just a placeholder for it, so it doesn't appear as blank.

What do you mean by half a div? Does this mean like it would be better if it was full-width (the width of select component?)?

ralph-dev:
> > You have made some nice radio components dual select and such. These are components in their own right and should be usable through the CE react ui library. I would export/make stories for them and put them in appropriate folders.
> 
> > 
> 
> > For the UI on both dual select radio dropdowns you could maybe use transitions to make the dropdown flow smoother. Look into the Select component/labellayout for how they implemented and use that.
> 
> > 
> 
> > Am i supposed to be able to interact with the upload/insert fields?
> 
> > 
> 
> > This dropdown only being half the div is odd to me, what do you think @ralph-dev ?
> 
> > 
> 
> > dropdown
> 
> 
> 
> Will do thanks for that advice.
> 
> 
> 
> The middle canvas is supposed to interact with other components, as in you should be able to drag things into these spaces, but the "upload image" is just a placeholder for it, so it doesn't appear as blank.
> 
> 
> 
> What do you mean by half a div? Does this mean like it would be better if it was full-width (the width of select component?)?
> 
> 
> 
> 

Hi Sharon,

He's just asking for your half width modal drop down to be the full width of the select for better UI

src/Containers/MiddleCanvas/DualSelectRadio.tsx
xsharonhe: I've tried to think of other ways to do so, but I think it's difficult since I need to control:
1. which text it changes in the select area (Flash Once/Beep Once) without the first line I would have no way to tell whether I should be changing flash or beep
2. the second and third one individually control the select options of the light and sound options and they do not rely on each other

Would you have any suggestions to reduce the lines of code here?

ralph-dev: Just put an onPress on the buttons, so you know which option is selected

ralph-dev: If you let the component trigger the action you will know which it is

src/Containers/MiddleCanvas/Template.tsx
```diff
@@ -0,0 +1,158 @@
...
+}): React.ReactElement => {
+    const [items, setItems] = useState<ITemplatePrefill[]>(Object.values(templatePrefills));
+
+    /**
+     * Reorders the draggable elements in a list
+     * @param {ITemplatePrefill[]} list - list of objects to reorder
+     * @param {number} startIndex - index of where the element originates from
+     * @param {number} endIndex - index of where the element will be placed
+     */
+    const reorder = (list: ITemplatePrefill[], startIndex: number, endIndex: number): ITemplatePrefill[] => {
```
xsharonhe, line 38: I need two different onDrag functions, but they are both very similar. I tried to export it and make it general, but it wouldn't really work with typing because list is of different type. Would it be a bad idea to make it an any type? If I make it an or type, it would still give me errors.

ralph-dev: Have you tried using the typescript
type command? That could solve this problem easily / clean up this code a bit. Your doing inline types for this function which is a bit messy

src/Containers/MiddleCanvas/RadioOptions.tsx
```diff
@@ -42,7 +34,12 @@ export const RadioOptions: React.FC<RadioOptionProps> = ({
                         name={title} 
                         label={label}
                         value={checkedOption === label}
-                        onChange={() => setCheckedOption(label)}
+                        onChange={() => {
+                            setCheckedOption(label);
+                            const newSelectOrder = [...selectOption];
+                            newSelectOrder[index] = label;
+                            setSelectOption(newSelectOrder);
```
Ampatte2, line 41: This should be a named function and not inline.

src/Containers/MiddleCanvas/RadioOptions.tsx
```diff
@@ -52,15 +49,15 @@ export const RadioOptions: React.FC<RadioOptionProps> = ({
...
-    font-stretch: normal;
-    font-style: normal;
     line-height: 1.25;
     letter-spacing: normal;
     padding: 5px 20px;
 `;
 const Text = styled.div`
     padding: 10px;
-    font-size: 16px;
+    ${({ theme }): string | undefined => `
```
Ampatte2, line 58: Same.

src/Containers/MiddleCanvas/Template.tsx
```diff
@@ -47,20 +33,18 @@ export const Template: React.FC<TemplateProps> = ({
...
     };
 
     /**
      * Returns a header component based on component type
-     * @param {componentType} componentType - type of component (like a table)
+     * @param {string | undefined} componentType - type of component (like a table)
      * @param {string[][]} labels - labels needed to be mapped in each individual component
      */
-    const conditionalHeaderComponent = (componentType: string, labels: string[][]): React.ReactElement | undefined => {
+    const conditionalHeaderComponent: getHeaderComponentType = (componentType, labels) => {
```
Ampatte2, line 47: For readability let's type the function rather than giving it a type. Ex: 
const conditionalHeaderComponent = (componentType: string | undefined, labels: string[ ][ ]): Reac.ReactElement =>

src/Containers/MiddleCanvas/Template.tsx
```diff
@@ -129,14 +113,12 @@ interface WrapperProps {
 };
 const Wrapper = styled.div<WrapperProps>`
     width: 364px;
```
Ampatte2, line 115: The static width/heights of the container elements will cause issues.

Ampatte2 (changes requested):
@ralph-dev is this component targeting tablet and greater or supposed to be usable at mobile. It currently breaks at small mobile, and large mobile has a x scroll.
Mobile-Small-Middle-Canvas
Mobile-Large-Middle-Canvas

ralph-dev:
> @ralph-dev is this component targeting tablet and greater only? It currently breaks at small mobile, and large mobile has a x scroll.
> 
> Mobile-Small-Middle-Canvas
> 
> Mobile-Large-Middle-Canvas
> 
> 
> 
> 

Tablet is the main focus but we should try to be responsive if possible in this case it's possible ( @xsharonhe can you use the scroll mixin and fix that ^

src/Containers/MiddleCanvas/DualSelectBar.tsx
```diff
@@ -17,12 +17,12 @@ export const DualSelectBar: React.FC<DualSelectBarProps> = ({
     setSelectedOption,
     ...props
 }): React.ReactElement => {
-    const getOptionFields = (field: string) => {
+    const getOptionFields = useCallback((field: string) => {
```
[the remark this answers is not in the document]
xsharonhe: I think this is what I did originally, but Ralph commented that I should avoid repetition in logic?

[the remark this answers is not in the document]
xsharonhe: I'll revert to what I had before, and see upon re-review what the general thoughts are because I do agree that this way seems more complicated

Ampatte2 (commented):
This is good, responsive, and readible. 

The draggable behavior is a bit difficult to use. Sometimes I click on an element and I can drag it, sometimes it drags the entire section.

Also, buisness information can not drag the first three elements, but I can drop into the first three elements.

VRBWA-132

ralph-dev:
@xsharonhe look at pointer events CSS and make sure you have it correctly. Or just add a triple bar drag icon to the Big Sections and have only that row be dragging the entire section rather than the entire Div

ralph-dev (commented):
@xsharonhe did you get to the comments yet?

src/Containers/MiddleCanvas/DroppableContainerContents.tsx
```diff
@@ -0,0 +1,35 @@
...
+            ))}
+        </Wrapper>
+    );
+}
+
+interface WrapperProps {
+    display?: boolean,
+};
+const Wrapper = styled.div<WrapperProps>`
+    ${({ display }): string => `
```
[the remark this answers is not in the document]
xsharonhe: If there is only one label, the label would be entered. If there is more than one label, the label would be spaced evenly. I think if I removed this the single label would just be on the left. 
Screen Shot 2020-11-28 at 4 51 15 PM

xsharonhe:
> This is good, responsive, and readible.
> 
> Needs a transition for the speed for the light/sound flash/beep drop down, it just pops open atm.
> 
> The draggable behavior is a bit difficult to use. Sometimes I click on an element and I can drag it, sometimes it drags the entire section.
> 
> Also, buisness information can not drag the first three elements, but I can drop into the first three elements.
> 
> VRBWA-132
> 
> Also @ralph-dev is this specific to storybook? Scroll bar does not have the scroll mixin styling.
> 
> Main Container Scroll

I'm not entirely sure if I used the transition mixin correctly? I've been trying to use it like the other examples, but it still is not giving me a very smooth transition.

src/Containers/MiddleCanvas/DroppableContainerContents.tsx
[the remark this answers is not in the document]
ralph-dev: Hmm I have a feeling that this works but the quality of code is weakened by it. There must be another css property that fits both cases, maybe just flex: 1 0: space between always?

src/Containers/MiddleCanvas/DroppableElement.tsx
```diff
@@ -0,0 +1,107 @@
...
+const DroppableContainer = styled.div<DroppableContainerProps>`
+    border-radius: 4px;
+    margin: 15px;
+    ${media(
+        'phone',
+        `
+        height: 60px;
+    `)};
+    ${({ theme, isDragging, isPreview }): string => `
+        background-color: ${isPreview ? theme.colors.background : theme.colors.input.default};
```
ralph-dev, line 101: You should group the set of isPreview CSS together instead of one line per each one btw

ralph-dev (approved):
I'm going to approve this but before we start another ticket I want to have a call to explain atomic design. I've drafted a very good tutorial (Better than any blog online, trust me I've checked) that explains it very well. I just need 10 minutes of your time :)

src/Containers/MiddleCanvas/DroppableContainerContents.tsx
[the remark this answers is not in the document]
xsharonhe: Not sure if this works out to be better?

ralph-dev (approved):
Looks better the styling!

PR #219, opened 2020-11-04
CEM-1189 Landing Component

Ampatte2:
Landing Component from Terminal and Dashboard.

Smalltext opacity 0.7 is odd for coloring, but it is the same method Info in LabelLayout uses, which is used by Input components.

CEM-1189-Landing-Component


ralph-dev (approved):
LGTM

PR #231, opened 2020-11-12
CEM-1214 Timepicker Close Outside Click

Ampatte2:
Timepicker's behavior for closing the list did not work when clicking on the parent element. 

Timepicker now closes when clicked outside of Timebox component rather than the parent.

CEM-1214 Timepicker-Close-Outside-Click


ralph-dev (approved):
LGTM, would be nice to see the cursor in the video so i know whats happening haha

PR #234, opened 2020-11-17
CEM-1234 update navigationitem

roggc:
Screenshot 2020-11-17 at 10 30 23
NavigationItem did not have exact property. This make that when a link other than home is selected, two links appears with active class, the one selected and home. With this fix then only one link would appear with active class, the one selected.


ralph-dev (approved):
Naming is a bit confusing but ok

PR #236, opened 2020-11-18
Cem 1231 SearchBar conversion from js to tsx

williancozzi:
I could not upload a pic or vid because something is breaking the project! I will send you the error @ slack.

But this pr worked fine, Andrew helped me and saw it working!


ralph-dev (approved):
Lgtm

PR #238, opened 2020-11-19
CEM-1239-Image-Component-Croping

Ampatte2:
This component did not function as intended. When cropping an image it would give you a distorted image that was off centered and did not return the correct crop.

Now the component returns the correctly cropped base64 image.

CEM-1239-Image-Component-Croping


ralph-dev (approved):
LGTM

PR #241, opened 2020-11-21
Conversion of Slider Story from js to tsx

williancozzi:
With two knobs. popups and showing values
image

Basic, with rail
image


williancozzi:
Just pushed the wrong branch.

PR #242, opened 2020-11-21
CEM-1242 converted Slider story from js to tsx

williancozzi:
With values, two knobs, rail and popups:
image

Basic with rail:
image


ralph-dev (approved):
LGTM

PR #243, opened 2020-11-23
CEM-1249 expand options when clicked on input

roggc:
combobox
Previous behavior did not expand options when clicking on the input field. Behavior still is not 100 percent satisfactory.


ralph-dev (approved):
LGTM

PR #244, opened 2020-11-24
CEM-1247 CustomSearch Story conversion from js to tsx

williancozzi:
Everything is fine except for this weird return when I changed food, but still show what is typed. Any thoughts on that?

Components-Custom-Search-Basic-⋅-Storybook-Google-Chrome-2020-11-23-22-29-31-1


ralph-dev (approved):
Lgtm,
It's possible the food field updates every text enter, and it didn't register the option click. It's ok your good to merge

PR #245, opened 2020-11-24
CEM-1251 Masked Input Fix

Ampatte2:
Reason for this fix, string coersion to number is bad. 

Also, updating onRealValueChange, more consistent to get an event from MaskedInput when dealing with other input elements. Specifically when dealing with external libraries such as formik where you do onRealValueChange={formik.handleChange} rather than having to setFieldValue name, value. This will also prove more extensible in the future.

CEM-1251-MaskedInput-Fix


ralph-dev (approved):
LGTM, prop name can just be onChange instead of onRealValuChange then lol but otherwise it's good

PR #247, opened 2020-11-26
Table managementv2 wait time

ScottyKnows44:


src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,98 @@
...
+ */
+function getColor(AverageWaitTime: Number){
+
+    const {colors} = useTheme();
+
+    if(AverageWaitTime >= 30){
+        return colors.PieChartColors.Red;
+    } else if(AverageWaitTime >= 15 && AverageWaitTime <= 30){
+        return 	colors.PieChartColors.Yellow;
+    } else {
```
ralph-dev, line 52: 30 is a magic number
const BUSY = 30;

ralph-dev (approved):
LGTM, please fix the magic numbers in the code

ralph-dev:
I should not see 15, 30 in the code becuase they have no meaning. const BUSY = 30, const STARTING_TO_GET_BUSY = 15 etc

src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+import styled, {useTheme} from "styled-components";
+
+export interface IWaitTimeDisplay {
+
```
ralph-dev, line 5: Why is there an extra line here?

src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,99 @@
...
+           AverageWaitTime = 29,
+            ...props
+       }) => {
+
+    return (
+        <Row {...props}>
+            <div>
+                <PieChart AverageWaitTime={AverageWaitTime} />
+            </div>
+
```
ralph-dev, line 27: Why is there random extra line heere?

src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,99 @@
...
+};
+
+
+/**
+ * This function will determine what color the Pie Chart will be
+ * @param AverageWaitTime {Number} - Number given from getTimeDifference() function
+ * @return {string} - Hexadecimal of color
+ */
+function getColor(AverageWaitTime: Number){
+    const BUSY = 30;
```
ralph-dev, line 45: Why are you re-initilizaing the variable BUSY everytime this function runs, does it change every function run? No, so make it a const outside of the function at the top of the file to keep things organized

src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const STARTING_TO_GET_BUSY = 15;
+    const {colors} = useTheme();
+
+    if(AverageWaitTime >= BUSY){
```
ralph-dev, line 49: use CSS to change CSS Styling not JavaScript, Move this logic into styled components

ralph-dev (approved):
LGTM

src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+const Row = styled.div`
+
+        display: flex;
+        flex-wrap: wrap;
+        margin-right: 15px;
+        margin-left: 15px;
+`;
+
+const PieChart = styled.div<Pick<IWaitTimeDisplay,"AverageWaitTime">>`
```
Ampatte2, line 85: Hmm. I think this isn't a problem that needs to be solved like this. Or if it is I think it can be solved more clearly with another interface. Currently the component only accepts one prop so this Pick doesn't perform a function. 

I would rather see PieChart have it's own interface as it's easier to add props to in the future if needed and cleaner code.

src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,99 @@
...
+const TextTime = styled.div`
+
+    font-size: 3rem;
+`;
+
+const TextFormatting = styled.div`
+
+    padding-left: 14rem;
+    padding-top: 2rem;
+`;
```
Ampatte2, line 75: So two things with these blocks of code here. First is that the naming is not very descriptive, it's just text and the variable. Name these more appropriately giving the reader more information about what they are.

Second, I think you can accomplish the same thing here with one of the text elements that we have in the React-UI library.

src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+const Row = styled.div`
+
+        display: flex;
+        flex-wrap: wrap;
```
Ampatte2, line 80: There is a mixin for this.

src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,99 @@
...
+        <Row {...props}>
+            <div>
+                <PieChart AverageWaitTime={AverageWaitTime} />
+            </div>
+
+            <TextFormatting>
+                <div>
+                    <TextAvgWait>Avg. Wait</TextAvgWait>
+                    <TextTime> {AverageWaitTime} Min</TextTime>
+                </div>
```
Ampatte2, line 32: What do these unstyled divs do?

src/Containers/WaitTimeDisplay.tsx
ralph-dev: @Ampatte2 It's better to use Pick then to redeclare the type that comes from the parent anyway for maintainability reasons. If the parent type changes, this will update the type automatically

src/Containers/SquareTable.tsx
```diff
@@ -86,6 +87,7 @@ export const SquareTable: React.FC<ISquareTable> = ({
     return (
         <div>
             {/** chairs top */}
+
```
ralph-dev, line 90: Why are there blank lines in the tree. The tree should be small anyways that blank lines are unessesary, lets remove these

src/Containers/SquareTable.tsx
```diff
@@ -68,6 +68,7 @@ export const SquareTable: React.FC<ISquareTable> = ({
      * @return {string} - Hexadecimal color value
      */
     function getOccupancyColor(): string {
+
```
ralph-dev, line 71: what's with the empty line?

src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,84 @@
...
+    AverageWaitTime: number,
+}
+
+/**
+ * Primary UI component for user interaction
+ * WaitTimeDisplay
+ */
+export const WaitTimeDisplay: React.FC<IWaitTimeDisplay>
+    = ({
+           AverageWaitTime = 29,
```
ralph-dev, line 22: 29 is a magic number, replace with const variable

src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * This function will determine what color the Pie Chart will be
+ * @param AverageWaitTime {Number} - Number given from getTimeDifference() function
+ * @return {string} - Hexadecimal of color
+ */
+function getColor(AverageWaitTime: Number){
+    const {colors} = useTheme();
+    if(AverageWaitTime >= BUSY){
+        return colors.PieChartColors.Red;
+    } else if(AverageWaitTime >= STARTING_TO_GET_BUSY && AverageWaitTime <= BUSY){
+        return 	colors.PieChartColors.Yellow;
```
ralph-dev, line 46: spacing :O

src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,84 @@
...
+    } else {
+        return colors.PieChartColors.Green;
+    }
+}
```
ralph-dev, line 50: This entire thing should be refactored into the Styled Component Location since this logic is assiociated with a specific component not the entire thing?

src/Containers/WaitTimeDisplay.tsx
```diff
@@ -0,0 +1,84 @@
...
+        flex-basis: 0;
+        max-width: 100%;
+        display: block; 
+        position: absolute; 
+        width: 10rem; 
+        height: 10rem; 
+        border-radius: 50%; 
+        
+        background-image: conic-gradient( 
+            ${ ({AverageWaitTime}) => getColor(AverageWaitTime) } ${({AverageWaitTime}) => AverageWaitTime * 6 }deg,
```
ralph-dev, line 81: 6 is a magic number

ralph-dev (commented):
Small Code Quality fixes left

src/Containers/WaitTimeDisplay.tsx
ScottyKnows44: Should I refractor this this into the PieChart component or move the function right above the component?

ralph-dev: I would move it next to the styled component becuase  it only relates to that component

ralph-dev (approved):
LGTM

PR #248, opened 2020-11-26
CEM-1250 Modal Story conversion from js to tsx

williancozzi:
Components-Modal-Basic-⋅-Storybook-Google-Chrome-2020-11-25-23-18-01-1


ralph-dev (approved):
Lgtm, not sure if those on clicks are being registered as actions as well but it's good

PR #250, opened 2020-11-29
converted typography story from js to tsx

williancozzi:
image

also just changed the extension and rendered fine.


ralph-dev (approved):
LGTM, I would have created a Typography Layout Component then just used that as the story haha but this works!

PR #257, opened 2020-12-05
CEM-1286-Vendors-Navigation-Bar-Molecule

xsharonhe:
I added a responsiveness feature from my last ticket (forgot about it).
Screen Shot 2020-12-05 at 6 09 01 PM

Navigation Bar:
Navigation Bar

Also would it be correct to call this an organism?

### DIFF
--- a/src/Containers/MiddleCanvas/DualSelectRadio.tsx
+++ b/src/Containers/MiddleCanvas/DualSelectRadio.tsx
@@ -147,15 +147,15 @@ interface SelectContainerProps {
 };
 const SelectContainer = styled.div<SelectContainerProps>`
     ${scroll};
-    ${({ theme, isVisible }): string => `
-        border: ${isVisible ? `solid 1px ${theme.colors.input.default}` : ''};
-        background-color: ${isVisible ? theme.colors.background : ''};
-        ${isVisible ? media(
+    width: 29%;
+    ${media(
         'tablet',
         `
-            width: 95%;
-            height: 220px;
-        `) : '180px'}; 
+            width: 95%
+        `)}; 
+    ${({ theme, isVisible }): string | false | undefined => isVisible &&`
+        border: 1px solid ${theme.colors.input.default};
+        background-color: ${theme.colors.background};
     `}
     border-radius: 8px;
     position: absolute;
--- a/src/Containers/VendorsList/NavigationBar.tsx
+++ b/src/Containers/VendorsList/NavigationBar.tsx
@@ -42,7 +42,7 @@ export const NavigationBar: React.FC<INavigationBarProps> = ({
 
 const Wrapper = styled.div`
     ${flex('row')};
-    ${({ theme }) => `
+    ${({ theme }): string => `
         border-bottom: 2px solid ${theme.colors.input.default};
     `};
     ${media(
--- a/src/Containers/VendorsList/NavigationBar.tsx
+++ b/src/Containers/VendorsList/NavigationBar.tsx
@@ -1,0 +1,54 @@
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { StyledIcon } from '@styled-icons/styled-icon';
+import { NavigationItem, INavigationItemProps } from './NavigationItem';
+import { MainInterface, ResponsiveInterface } from '../../Utils/BaseStyles';
+import { flex, media, scroll } from '../../Utils/Mixins';
+
+
+export interface INavigationBarItems {
+    icon?: StyledIcon;
+    label: string
+};
+
+export interface INavigationBarProps extends MainInterface, ResponsiveInterface, React.HTMLAttributes<HTMLDivElement> {
+    navigationBarItems: INavigationBarItems[];
+    navigationItemProps?: INavigationItemProps;
+};
+
+const FIRST_LABEL = 0;
+
+export const NavigationBar: React.FC<INavigationBarProps> = ({
+    navigationBarItems,
+    navigationItemProps,
+    ...props
+}): React.ReactElement => {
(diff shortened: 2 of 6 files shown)


src/Containers/VendorsList/NavigationBar.tsx
```diff
@@ -0,0 +1,54 @@
...
+};
+
+const FIRST_LABEL = 0;
+
+export const NavigationBar: React.FC<INavigationBarProps> = ({
+    navigationBarItems,
+    navigationItemProps,
+    ...props
+}): React.ReactElement => {
+    const [selectedItem, setSelectedItem] = useState(navigationBarItems[FIRST_LABEL].label);
```
ralph-dev, line 26: Uncaught TypeError: Cannot read property 'label' of undefined at <anonymous>:1:7

Edge missing maybe? ?.label

src/Containers/VendorsList/NavigationBar.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { StyledIcon } from '@styled-icons/styled-icon';
+import { NavigationItem, INavigationItemProps } from './NavigationItem';
+import { MainInterface, ResponsiveInterface } from '../../Utils/BaseStyles';
+import { flex, media, scroll } from '../../Utils/Mixins';
+
+
+export interface INavigationBarItems {
```
ralph-dev, line 9: Why aren't you re-using the type of NavigationBar itself? not sure why creating a new interface here?

src/Containers/VendorsList/NavigationItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+    isSelected: boolean
+};
+const Wrapper = styled.div<IWrapperProps>`
+    ${flex('row')};
+    ${({ theme, isSelected }): string => `
+        border-bottom: ${isSelected ? `solid 4px ${theme.colors.primary}` : ''};
+    `};
+`;
+
+interface IIconProps {};
```
ralph-dev, line 44: You can remove this

src/Containers/VendorsList/NavigationItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+                {label}
+            </Heading>
+        </Wrapper>
+    );
+};
+
+interface IWrapperProps{
+    isSelected: boolean
+};
+const Wrapper = styled.div<IWrapperProps>`
```
ralph-dev, line 37: You could also do Pick<INavigationItemProps, 'selectedItem'> then pass selectedItem down and do the comparison in the Styled Component as well. That would keep the types in sync / reduce extra code being written

ralph-dev: Love to hear your comment on this though ^

src/Containers/VendorsList/NavigationItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+    ${flex('row')};
+    ${({ theme, isSelected }): string => `
+        border-bottom: ${isSelected ? `solid 4px ${theme.colors.primary}` : ''};
```
ralph-dev, line 40: does
isSlected && Styling work or not really? if it doesn't this is fine

ralph-dev (approved):
LGTM have you made sure that the controls are showing? Great catch with the responsive point!

ralph-dev:
Oh yeah, why is this an organism btw?
3 Buttons = 3 Atoms,
1 Container = 1 Atoms, Thus it's a molecule?

xsharonhe:
> Oh yeah, why is this an organism btw?
> 3 Buttons = 3 Atoms,
> 1 Container = 1 Atoms, Thus it's a molecule?

Yeah, I had a feeling after the last component that this was a molecule, since this was a relatively small component. I saw the svg and heading as a "molecule" in itself because I built them together, but thinking about it as a bigger picture, it should be a molecule.

xsharonhe:
> LGTM have you made sure that the controls are showing? Great catch with the responsive point!

Yup, made sure controls worked this time haha

src/Containers/VendorsList/NavigationItem.tsx
xsharonhe: Interesting, never thought of this method, but I do agree that it is cleaner / saves lines of code. I have to pass in two props though since it relies on a boolean value? Not sure if this still works well

ralph-dev (approved):
Lgtm

src/Containers/MiddleCanvas/DualSelectRadio.tsx
```diff
@@ -147,15 +147,15 @@ interface SelectContainerProps {
...
+    width: 29%;
+    ${media(
         'tablet',
         `
-            width: 95%;
-            height: 220px;
-        `) : '180px'}; 
+            width: 95%
+        `)}; 
+    ${({ theme, isVisible }): string | false | undefined => isVisible &&`
```
ralph-dev, line 156: Curious how this would be undefined? otherwise ok

src/Containers/MiddleCanvas/DualSelectRadio.tsx
```diff
@@ -147,15 +147,15 @@ interface SelectContainerProps {
...
-        ${isVisible ? media(
+    width: 29%;
+    ${media(
         'tablet',
         `
-            width: 95%;
-            height: 220px;
-        `) : '180px'}; 
+            width: 95%
+        `)};
```
ralph-dev, line 155: A bit weird you didn't just say
width 95%
then only on Phone make the width smaller?

Becuase logically,
Tablet, Desktop Everthing Else is width 95% and only on phone is width 29%.

PR #259, opened 2020-12-11
CEM-1291 Email, name, profile molecule

xsharonhe:
Bigger pull request than I anticipated, so I'll get an approval on this molecule and move on to the table. 
Screen Shot 2020-12-11 at 4 41 12 PM

### DIFF
--- a/src/Containers/VendorsList/Profile.tsx
+++ b/src/Containers/VendorsList/Profile.tsx
@@ -6,27 +6,76 @@ import { MainInterface, ResponsiveInterface } from '../../Utils/BaseStyles';
 import { flex, media } from '../../Utils/Mixins';
 
 export interface IProfileProps extends MainInterface, ResponsiveInterface,React.HTMLAttributes<HTMLDivElement> {
-    headerRowProps?: HeaderRowProps
+    headerRowProps?: HeaderRowProps;
+    imageUrl?: string;
+    key: number;
+    name: string;
+    email: string;
+    profileProps?: IFeaturedProfileProps;
 }
 
+const MATCH_FIRST_LETTER = /\b\w/g;
+
 export const Profile: React.FC<IProfileProps> = ({
+    imageUrl, 
+    name, 
+    email,
+    key,
     headerRowProps,
+    profileProps,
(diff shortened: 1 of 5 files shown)


src/Containers/VendorsList/Profile.tsx
```diff
@@ -0,0 +1,91 @@
...
+     */
+    const getProfiles = (inputName: string, image?: string) => {
+        const profileInitials = getInitials(inputName);
+
+        switch(true) {
+        case image === undefined:
+            return (
+                <FeaturedProfile 
+                    {...profileProps}
+                    background='orange'
```
xsharonhe, line 51: is there any way to use theme here? background is a mandatory prop as a string

ralph-dev: const theme = useTheme();
Styled components hook.

Then theme.Colors.Etc

src/Containers/VendorsList/Profile.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { FeaturedProfile, IFeaturedProfileProps } from '../FeaturedProfile';
+import { HeaderRow, HeaderRowProps } from '../HeaderRow';
+import { MainInterface, ResponsiveInterface } from '../../Utils/BaseStyles';
+import { flex, media } from '../../Utils/Mixins';
+
+export interface IProfileProps extends MainInterface, ResponsiveInterface,React.HTMLAttributes<HTMLDivElement> {
+    headerRowProps?: HeaderRowProps;
+    imageUrl?: string;
+    key: number;
+    name: string;
```
ralph-dev, line 12: This name and email and URL typing is repeated, why not just extend the other interface? No duplicate code.

ralph-dev (approved):
Lgtm

src/Containers/VendorsList/Profile.tsx
xsharonhe: Would it be recommended to move interfaces to another .ts file because if I extend it, it would give me a dependency error, since I'm giving it the options to extend props to the profile component already?

PR #260, opened 2020-12-11
CEM-1215-Phone-Number-Mask

Ampatte2:
MaskedInput did not support exterior error handling, and for errors to display required a min/max value. Now supports exterior error handling and default error handling.

Phone number mask has been added. Will still error for invalid input, but on blur will now display a formatted phone number with the format 1-xxx-xxx-xxxx. 
CEM-1215-Phone-Number-Mask - Mask Update
CEM-1215-Phone-Number-Mask


ralph-dev (approved):
Lgtm

PR #261, opened 2020-12-11
CEM-773 SegmentedButton

Ampatte2:
A segmented button that takes an array of segment types and creates a row of buttons.

CEM-773-SegmentedButton

CEM-773-SegmentedButton - More
CEM-773-SegmentedButton - Less

### DIFF
--- a/src/Inputs/SegmentedButton.tsx
+++ b/src/Inputs/SegmentedButton.tsx
@@ -19,7 +19,7 @@ export interface ISegmentedButtonProps extends Omit<React.HTMLAttributes<HTMLDiv
 
 /**
 *@type  {ISegment[]} 
-*@param {segments} - array of object with a name and value active either true or false.
+*@param {segments} - objects with a name, active boolean, and icon properties;
 * */
 
 export const SegmentedButton : React.FC<ISegmentedButtonProps> = ({
@@ -63,7 +63,6 @@ export const SegmentedButton : React.FC<ISegmentedButtonProps> = ({
 interface ISegmentProps extends React.HTMLAttributes<HTMLDivElement> {
     active?: boolean;
 }
-
 interface ISegmentContainerProps extends React.HTMLAttributes<HTMLDivElement> {
     width?: string;
     height?: string;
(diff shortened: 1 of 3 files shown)


ralph-dev (approved):
LGTM, a bit weird you override the onClick but must be for good reason?

src/Inputs/SegmentedButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+    onClick,
+    ...props
+}): React.ReactElement => {
+
+    const renderSegments = useCallback(() => {
+        return segments.map((segment, index)=> {
+            return (
+                <Segment 
+                    active={segment.active}
+                    onClick={(event)=> onClick(event, index)}
```
ralph-dev, line 39:
```suggestion
                    onClick={(event) => onClick(event, index)}
```

src/Inputs/SegmentedButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+        height: ${height};
+    `}
+`
+const Segment = styled.div<ISegmentProps>`
+    cursor: pointer;
+    ${transition(['background-color', 'opacity'])}
+    ${flex('center')}
+    ${clickable('#ffffff', 0.05)};
+    ${({theme, active})=> {
+        const style = active ? `background-color: ${theme.colors.primary}; color: ${theme.colors.background};` : ` background-color: ${theme.colors.background}; color: ${theme.colors.text};`
```
ralph-dev, line 94: what is this lmfao

ralph-dev: Have you tried 
CSS Variants :)

Ampatte2:
Wanted to expose the index of the array, and in the past have learned that having access to the event information is always a plus.

src/Inputs/SegmentedButton.tsx
ralph-dev: image

Ampatte2: I thought was originally changing clickable, and have had troubles with interpolating the mixins in the past. I will revert to just the ternary

PR #263, opened 2020-12-12
CEM-1287 Vendors-Table-Organism

xsharonhe:
Changed the tag file so that it is optional to hover over because this use case does not need the x icon.

Breaks at small phone (must scroll horizontally to see it fully), but works at all other media points. 
Screen Shot 2020-12-14 at 8 12 56 PM
Screen Shot 2020-12-14 at 8 13 13 PM


stories/Containers/VendorsTable.stories.tsx
```diff
@@ -8,20 +15,92 @@ export default {
...
+                    email={cell.row.original.email} 
+                    imageUrl={cell.row.original?.imageUrl} 
+                />
+            )
+        },
+        {
+            Header: 'Tags',
+            accessor: 'col_tags',
+            Cell: cell => (
+                <TagContainer
```
xsharonhe, line 39: Had questions on how to access the props' style. For instance, I need to add margin between the props and I need iconProps' style, or is there a method for doing so?

src/Containers/Tag.tsx
```diff
@@ -6,18 +6,20 @@ import { Main } from '../Utils/BaseStyles';
...
     icon = Times,
     children,
+    hasHoverMode = true,
     ...props
 }): React.ReactElement => {
     return (
         <TagDiv {...props}>
             {children}
-            <Icon as={icon} />
+            {!!hasHoverMode && <Icon as={icon} />}
```
ralph-dev, line 22: This code is a bit confusing, the icon only shows if what has hoverMode, the tag? What does the icon has to do with the styling has Hover mode?
Also, isHoverable is a better name ;)

src/Containers/VendorsList/Profile.tsx
```diff
@@ -15,6 +15,8 @@ export interface IProfileProps extends MainInterface, ResponsiveInterface,React.
 }
 
 const MATCH_FIRST_LETTER = /\b\w/g;
+const WIDTH = 50;
```
ralph-dev, line 18: PROFILE_WIDTH, width is very broad could apply to anything

stories/Containers/VendorsTable.stories.tsx
ralph-dev: What do you mean here, I would like this to be explained a bit clearer a call might be helpful as well.

ralph-dev (commented):
Small comments

src/Containers/ReactTable.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { MainInterface, ResponsiveInterface } from '../Utils/BaseStyles';
+import { media, scroll } from '../Utils/Mixins';
+
+export interface IVendorsData extends IProfileProps {
+    tags?: string[];
+    createdAt?: string;
+};
+
+export interface IReactTableProps extends MainInterface, ResponsiveInterface,React.HTMLAttributes<HTMLDivElement> {
+    data: IVendorsData[];
```
xsharonhe, line 14: This component should be fully reusable for any component besides the typing. Is there any way to pass in a general interface here? or is that the type any?

ralph-dev: You could do object[]
But the best thing to do is most likely involving typescript generics :)

src/Containers/ReactTable.tsx
```diff
@@ -10,14 +10,14 @@ export interface IVendorsData extends IProfileProps {
...
-    columns: Column<IVendorsData>[];
+export interface IReactTableProps<T extends IVendorsData> extends MainInterface, ResponsiveInterface,React.HTMLAttributes<HTMLDivElement> {
+    data: T[];
+    columns: Column<T>[];
     tableHeaderProps?: TableHeaderProps;
     tableRowProps?: TableRowProps;
 };
 
-export const ReactTable: React.FC<IReactTableProps> = ({
+export const ReactTable: React.FC<IReactTableProps<any>> = ({
```
ralph-dev, line 20: This should be T not any?

src/Containers/ReactTable.tsx
```diff
@@ -0,0 +1,114 @@
...
+    return (
+        <table {...getTableProps()} {...props}>
+            <STableHead>
+                {headerGroups.map((headerGroup, index) => (
+                    <SHeadTableRow {...headerGroup.getHeaderGroupProps()} key={headerGroup.headers[index].Header?.toString()}>
+                        {headerGroup.headers.map(column => (
+                            <STableHeader {...column.getHeaderProps()} key={column.Header?.toString()} {...tableHeaderProps}>
+                                {column.render('Header')}
+                            </STableHeader>
+                        ))}
```
ralph-dev, line 47: It's best to keep the tree small and make a function called getColumns, that way you can wrap it with a useCallback / Memo if needed

src/Containers/ReactTable.tsx
```diff
@@ -0,0 +1,114 @@
...
+                    return (
+                        <STableRow {...row.getRowProps()} key={row.original.id} {...tableRowProps}>
+                            {row.cells.map(cell => (
+                                <STableData {...cell.getCellProps()}>
+                                    {cell.render('Cell')}
+                                </STableData>
+                            ))}
+                        </STableRow>
+                    );
+                })}
```
ralph-dev, line 63: Same here, inline logic isn't very good practice usually

src/Containers/VendorsList/Profile.tsx
```diff
@@ -5,16 +5,18 @@ import { HeaderRow, HeaderRowProps } from '../HeaderRow';
...
+export interface IProfileProps extends MainInterface, ResponsiveInterface, React.HTMLAttributes<HTMLDivElement> {
     key: number;
-    name: string;
-    email: string;
+    name?: string;
+    email?: string;
     imageUrl?: string;
     headerRowProps?: HeaderRowProps;
-    profileProps?: IFeaturedProfileProps;
+    profileProps?: Omit<IFeaturedProfileProps, 'key, background'>;
```
ralph-dev, line 14: shouldn't this be 'key' , 'background'? if this works i am probs confusing the syntax and you are correct lol

src/Containers/VendorsList/Profile.tsx
```diff
@@ -49,19 +51,23 @@ export const Profile: React.FC<IProfileProps> = ({
...
             return (
                 <FeaturedProfile 
-                    {...profileProps}
                     image={image}
                     background='none'
                     key={key} 
+                    width={PROFILE_WIDTH}
+                    height={PROFILE_HEIGHT}
+                    {...profileProps}
                 />
```
ralph-dev, line 71: There are duplicate props here, abstract the props into an object then modify that object and spread it across both of these / just have one return statement. lots of cleaner code can be done here

src/Containers/VendorsList/TagContainer.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Tag, TagProps } from '../Tag';
+import { MainInterface, ResponsiveInterface } from '../../Utils/BaseStyles';
+import { flex, media } from '../../Utils/Mixins';
+
+export interface ITagContainerProps extends MainInterface, ResponsiveInterface, React.HTMLAttributes<HTMLDivElement> {
+    tags: string[];
+    isHoverable?: boolean;
+    tagProps?: Omit<TagProps, 'children'>;
```
ralph-dev, line 10: I would explore why children was added as a prop in the first place btw, otherwise this is good

ralph-dev (commented):
Some feedback on code quality

src/Containers/ReactTable.tsx
xsharonhe: T gave me an error but I think I did it incorrectly, so I made the change

src/Containers/VendorsList/Profile.tsx
xsharonhe: Oops yeah haha I did do it wrong thanks for the catch

src/Containers/VendorsList/Profile.tsx
```diff
@@ -40,45 +42,41 @@ export const Profile: React.FC<IProfileProps> = ({
...
-                    background='none'
-                    key={key} 
-                />
-            );
-        }
+        const getDefaultProfileProps = ({
+            key: id,
+            width: PROFILE_WIDTH,
+            height: PROFILE_HEIGHT,
+            background: image === undefined ? theme.colors.primary : 'none',
```
ralph-dev, line 54: what if image is nulll?
Why not use !!image

ralph-dev (approved):
LGTM, will let Andrew take a look when he has time

PR #267, opened 2020-12-21
CEM-1288-Pagination-Molecule

xsharonhe:
Sorry for the large PR. Most of it was just repetitive addition of data in stories.
Pagination-Molecule


ralph-dev (commented):
Looks good so far, constants are all caps btw

ralph-dev (approved):
LGTM, let me know when you are almost done as I want to have a call to discuss some new topics / prep you for the next concept. Since it's the holidays we can have the call whenever you like maybe the monday - wednesday

PR #269, opened 2020-12-23
CEM-1321 fileupload component

roggc:
https://drive.google.com/file/d/1IiAAxznLh5FHcGy2kxbG4shqD2bMA1h2/view?usp=sharing
disabled
dragenteranimation
disabledeffect


src/Containers/FileUpload/FileUpload.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+import {Image} from '@styled-icons/fa-solid/Image';
+import {TextLayout} from '@Layouts';
+import styled from 'styled-components'
+import { FileUpload,IFileUploadProps } from '../../index';
+import { createStoryTitle } from '../../Constants';
+
+const A=styled.a`
```
ralph-dev, line 9: Bottom of the file

ralph-dev (commented):
Lgtm, now onto event handlers

src/Containers/FileUpload/FileUpload.stories.tsx
roggc: no, because if not error about using a variable before declaring appears.

ralph-dev (commented):
Looks really good, where is the success animation?

ralph-dev:
Ok the component is looking better, but what about these cases 
1) When the file is successfully loaded does it appear anywhere?
2) If the Image Picker allows for only 1 file to be uploaded max how does the UI look for replacement.
3) If it allows more than one file how does it show a file that's been uploaded / allow you to upload more UI

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,267 @@
...
+    isUploading,
+    isSuccess,
+    isFailure,
+    successMessage,
+    failureMessage,
+}):React.ReactElement=>{
+    const [height,setHeight]=useState<number|undefined>(undefined)
+    const [totalHeight,setTotalHeight]=useState(0)
+    const [totalHeightPlus,setTotalHeightPlus]=useState(0)
+    const [padding]=useState(10)
```
ralph-dev, line 38: Why is this here? If your using this for conditional animation it should be in styled components. If your using this for rendering an animations you should use CSS Keyframes

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,267 @@
...
+    const [maxHeight,setMaxHeight]=useState<number|undefined>(undefined)
+    const [positionLoading,setPositionLoading]=useState(!isUploading)
+    const [positionTopLoading,setPositionTopLoading]=useState(0)
+    const [opacityLoading,setOpacityLoading]=useState(0)
+    const [positionIsSuccess,setPositionIsSuccess]=useState(!isSuccess)
+    const [opacityIsSuccess,setOpacityIsSuccess]=useState(0)
+    const [loadingContainerHeight,setLoadingContainerHeight]=useState(0)
+    const [widthComponent,setWidthComponent]=useState<number>(0)
+    const [margin]=useState(10)
```
ralph-dev, line 47: No

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,267 @@
...
+    const [totalHeightPlus,setTotalHeightPlus]=useState(0)
+    const [padding]=useState(10)
+    const [maxHeight,setMaxHeight]=useState<number|undefined>(undefined)
+    const [positionLoading,setPositionLoading]=useState(!isUploading)
+    const [positionTopLoading,setPositionTopLoading]=useState(0)
+    const [opacityLoading,setOpacityLoading]=useState(0)
+    const [positionIsSuccess,setPositionIsSuccess]=useState(!isSuccess)
+    const [opacityIsSuccess,setOpacityIsSuccess]=useState(0)
+    const [loadingContainerHeight,setLoadingContainerHeight]=useState(0)
+    const [widthComponent,setWidthComponent]=useState<number>(0)
```
ralph-dev, line 46: Do you need to specify the number type?

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,267 @@
...
+            setPositionIsSuccess(true)
+        }
+        return ()=>{
+            setPositionIsSuccess(false)
+            setOpacityIsSuccess(1)
+        }
+    },[isSuccess,isFailure])
+
+
+    useEffect(()=>{
```
ralph-dev, line 120: Write descriptions for these use effects you have so many

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,267 @@
...
+            <Container withBorder padding='30px 20px 43px 20px' opacity={opacityLoading} position={positionLoading} ref={loadingContainerRef} overflow='hidden' width={widthComponent} margin={`${margin}px`} positionTop={positionTopLoading}>
+                <Loading loading={isUploading} message='Uploading...' />
+            </Container>
+            <Container withBorder opacity={opacityIsSuccess} height={loadingContainerHeight} position={positionIsSuccess} margin={`${margin}px`} positionTop={positionTopLoading} width={widthIsSuccess}>
+                <Container withFlexSpaceBetween>
+                    <TextLayout bold color='DarkBlue'>{successMessage}</TextLayout>
+                    <Icon as={CheckCircle} color={MainTheme.colors.statusColors.green} />
+                </Container>
+            </Container>
+            <Container withBorder opacity={opacityIsFailure} height={loadingContainerHeight} position={positionIsFailure} margin={`${margin}px`} positionTop={positionTopLoading} width={widthIsSuccess}>
```
ralph-dev, line 193: These should be sub components, becuase they are only used under this component

ralph-dev (changes requested):
image

src/Containers/FileUpload/FileUpload.tsx
roggc: I use this padding state variable to calculate (compute) the width of the element and set it when its position is absolute.

src/Containers/FileUpload/FileUpload.tsx
roggc: no

src/Containers/FileUpload/FileUpload.tsx
ralph-dev: I think this is incorrect, if you can explain more about why it's being used / how that will help suggest a better solution. Are you sure you don't need a useRef or anything else I'm sure there is a better way

src/Containers/FileUpload/BottomPanel.tsx
```diff
@@ -0,0 +1,30 @@
+import React,{forwardRef} from 'react'
+import {Container} from './Container'
+
+interface IBottomPanelProps{
```
ralph-dev, line 4: Why is this not extending HTMLDivElemt so all the props don't need to be retyped??

src/Containers/FileUpload/Container.ts
```diff
@@ -0,0 +1,43 @@
+import styled from 'styled-components'
+import {flex} from '@Utils/Mixins'
+
+interface IContainerProps{
```
ralph-dev, line 4: not extending correct interface

ralph-dev (changes requested):
What is this code quality, please ask questions about TypeScript if you don't understand something.

src/Containers/FileUpload/BottomPanel.tsx
```diff
@@ -0,0 +1,13 @@
+import React,{forwardRef} from 'react'
+import {Container,IContainerProps} from './Container'
+
+interface IBottomPanelProps extends IContainerProps{
+}
+
+export const BottomPanel=forwardRef<HTMLDivElement,IBottomPanelProps>(({children,...props},forwadedRef):React.ReactElement=>{
```
ralph-dev, line 7: What's by is there a <> here

src/Containers/FileUpload/Container.ts
```diff
@@ -0,0 +1,43 @@
+import styled from 'styled-components'
+import {flex} from '@Utils/Mixins'
+
+export interface IContainerProps{
```
ralph-dev, line 4: Why not extend Div?

ralph-dev (changes requested):

src/Containers/FileUpload/BottomPanel.tsx
roggc: I don't understand sorry

src/Containers/FileUpload/Container.ts
roggc: I think it's not necessary in the case of styled components.

src/Containers/FileUpload/BottomPanel.tsx
ralph-dev: Why do you have <Div, props>
Shouldn't the props just extend Div?

roggc: This is how it is done when you need to forward a ref in typescript

src/Containers/FileUpload/Container.ts
ralph-dev: But then why are you redeclaring are the styled props?

If you want all the style props exposed without doing this manually and having them optional why not just do

Partial<Pick<Div, Style>>

ralph-dev (changes requested):
You component is missing the animation for onHover btw
I want you to integrate Lottie in this PR, it's very easy and will make your components higher quality, https://lottiefiles.com/search?q=Files&category=animations

src/Containers/FileUpload/BottomPanel.tsx
ralph-dev: But why do you need Div, Props
Can you just have Props
Then have props extend Div?

ralph-dev (changes requested):
File Upload / Container File Interface what is that??
Everytime I find a mistake in the PR I am going to add an additional UI / UX requirement until your code is perfect 👍🏻 
New Requirement: Let's have the Dashboard Border Animates (As in Moves around like a train) so that the dashed lines are animating when you've picked a file up, like the animation :)

src/Containers/FileUpload/BottomPanel.tsx
roggc: HTMLDivElement refers to the ref type

src/Containers/FileUpload/Container.ts
roggc: Sorry, I don't get it. What is Style? I have personal options (properties) defined by me which I need. That's why I declare them.

src/Containers/FileUpload/BottomPanel.tsx
ralph-dev: Then isn't the ordering incorrect?
Becuase the ref is the second parameter not the first

src/Containers/FileUpload/Container.ts
ralph-dev: I do not understand what is happening here at all. Can you try to explain what EXACTLY you are trying to do here.
DragEnter comes from https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.dragevent.html 
DragEvent, i don't know why you are creating custom types? Margin is already from div elements etc

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,236 @@
...
+        return ()=>{
+            setPositionLoading(false)
+            setOpacityLoading(1)
+        }
+    },[isUploading,isSuccess,isFailure])
+
+    // this is to fade out success container component
+    useEffect(()=>{
+        if(!isSuccess||isFailure){
+            setOpacityIsSuccess(0)
```
ralph-dev, line 120: Why is opacity have loading / success states? Can the naming be improved

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,236 @@
...
+                <Container withFlexSpaceBetween>
+                    <TextLayout bold color='DarkBlue'>{successMessage}</TextLayout>
+                    <Icon as={CheckCircle} color={MainTheme.colors.statusColors.green} />
+                </Container>
+            </BottomPanel>
+            <BottomPanel withBorder opacity={opacityIsFailure} height={loadingContainerHeight} position={positionIsFailure} margin={`${margin}px`} positionTop={positionTopLoading} width={widthIsSuccess}>
+                <Container withFlexSpaceBetween>
+                    <TextLayout bold color='DarkBlue'>{failureMessage}</TextLayout>
+                    <Icon as={TimesCircle} color={MainTheme.colors.statusColors.red} />
+                </Container>
```
ralph-dev, line 207: Why aren't you re-using code????

ralph-dev (changes requested):
Getting better still work left to do

src/Containers/FileUpload/FileUpload.tsx
roggc: because there is opacity for the isUploading panel, opacity for the isSuccess panel and opacity for the isFailure panel

src/Containers/FileUpload/FileUpload.tsx
roggc: I don't know how I can reuse code in this case. BottomPanel renders it's children, which in this case there are three different bottom panel with three different children (there are two similar, but one different, which is the isUploading bottompanel)

ralph-dev: Children as a prop then?

ralph-dev: Just inject the things that are different as props

ralph-dev:
New requirement: When in Disbaled State, UI should have opacity faded or look disabled!

src/Containers/FileUpload/FileUpload.tsx
roggc: Yes, that's it. children it's already a prop. The only thing that occurs to me is to make a react component for the two similar children.

ralph-dev:
@roggc Which gif contains this case?
Max One File, With One File Already Uploaded (Thus Disabled becuase max files reached?)
Also fix merge conflicts

roggc:
@ralph-dev , the last gif where the disabled property is turned on should be the case where no more files are allowed.

src/Containers/FileUpload/BottomPanel.tsx
roggc: Yes, that's right, but this is how it is. blame react! hehe

src/Containers/FileUpload/Container.ts
roggc: Is not DrageEnter, is isDragEnter, which is a boolean. and margin is not a valid div attribute, as far as I know

Ampatte2 (commented):
I see a lot of useState and useEffect. This code is difficult to understand and read, I am thinking this may be better with a reducer for a few reasons. 

Clarity of the naming/variables. Instead of 10 different states now you can organize variables within your single state such as creating properties like loading { opacity: 0, position: 0}, isSuccess {position: 0, opacity: 0}. 

Instead of multiple useEffects to update you can have one useEffect trigger an action based on conditions. And that action is named to give the reader an idea of what those conditions mean together.

For something as complicated as this component I think being verbose is definitely the way to go, if I needed to debug this in the current state I would need to learn the entire thing.

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -38,135 +188,135 @@ export const FileUpload:React.FC<IFileUploadProps>=({
...
+    },[isUploading,isSuccess,isFailure,state.totalHeight,state.totalHeightPlus])
 
     // this is to calculate (set) some values
     useEffect(()=>{
         if(containerRef.current?.scrollHeight){
             const contentHeight=containerRef.current.scrollHeight
-            setTotalHeight(contentHeight-padding*2)
-            setMaxHeight(contentHeight-padding*2)
+            dispatch({type:SET_TOTALHEIGHT,value:contentHeight-state.padding*2})
+            dispatch({type:SET_MAXHEIGHT,value:contentHeight-state.padding*2})
```
Ampatte2, line 228: These can be combined into one dispatch, and the same for line 215.

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -38,135 +188,135 @@ export const FileUpload:React.FC<IFileUploadProps>=({
...
         }
         if(rootRef.current?.clientWidth){
-            setWidthComponent(rootRef.current?.clientWidth-margin*2-padding*2)
+            dispatch({type:SET_COMPONENTWIDTH,value:rootRef.current.clientWidth-state.margin*2-state.padding*2})
```
Ampatte2, line 231: I am not the biggest fan of computing values within setting a state or dispatching, declaring your computed value with a named variable then passing that into the dispatch/state gives more information about what is happening.

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -38,135 +188,135 @@ export const FileUpload:React.FC<IFileUploadProps>=({
...
-    },[isSuccess,isFailure])
+    },[isSuccess,isFailure,state.padding,state.margin])
 
     // this is to fade out uploading container component
     useEffect(()=>{
         if(!isUploading||isSuccess||isFailure){
-            setOpacityLoading(0)
-            setPositionLoading(true)
+            dispatch({type:SET_OPACITYLOADING,value:0})
+            dispatch({type:SET_POSITIONLOADING,value:true})
```
Ampatte2, line 261: Combine and name.

Ampatte2 (commented):
This mainly converts states into actions, which doesn't solve the issue that it is complicated. There are some dispatches which are called together and can be combined.

There are five useEffects that have similiar dependancies of isSuccess, isFailure, and isUploading (in various combinations), is it possible to combine these and handle certain actions based on common conditions? (Let me know what you think).

After combining the actions is there a way to better name the them to tell the reader what change this is effecting or why this action is changing the parts of the state it is?

src/Containers/FileUpload/FileUpload.tsx
roggc: how do you combine it into only one dispatch?

src/Containers/FileUpload/FileUpload.tsx
roggc: how do you combine?

roggc: ok, I know it. Let me try it...

src/Containers/FileUpload/FileUpload.tsx
roggc: yes, I know it. I will do it...

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -157,6 +165,60 @@ const reducer=(state:IState,action:Action):IState=>{
...
+        }
+    case ISFAILURE_RESTORE:
+        return {
+            ...state,
+            isFailure:{
+                ...state.isFailure,
+                opacity:1,
+                position:false,
+            },
+        }
```
Ampatte2, line 221: Yes this is much better! I have a better understanding of why this action is taking place!

Ampatte2 (commented):
There are other dispatches like line 277 and 289 that can be combined. Still needs some actions to be renamed to be more descriptive. The convention of ISSUCCESS_FADEOUT would be a boolean, SUCCESS_FADEOUT is sufficient. 

Let's try to condense the actions more and combine useEffects that are dependant on isSuccess, isFailure, and isUploading.

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -350,7 +350,8 @@ export const FileUpload:React.FC<IFileUploadProps>=({
     useLayoutEffect(() => {
         function updateSize() {
             if(rootRef.current?.clientWidth){
-                dispatch({type:SET_COMPONENTWIDTH,value:rootRef.current.clientWidth-state.margin*2-state.padding*2})
+                const width=rootRef.current.clientWidth-state.margin*2-state.padding*2
```
Ampatte2, line 353: Needs better naming, what is this calculation performing? Descriptive is better than short.

src/Containers/FileUpload/BottomPanel.tsx
ralph-dev: Oh my god you are right haha this is so inconsistent with everything else, god damn React!!

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,451 @@
...
+        return {
+            ...state,
+            totalHeightPlus:action.value
+        }
+    case SET_POSITIONTOPLOADING:
+        return {
+            ...state,
+            positionTopLoading:action.value
+        }
+    case SET_ISSUCCESSWIDTH:
```
ralph-dev, line 107: SET_IS_SUCCESS_WIDTH

SCREAMING_SNAKE_CASE

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,456 @@
...
+    isSuccess:IOptions;
+    isFailure:IOptions;
+    positionTopLoading:number;
+    loadingContainerHeight:number;
+    componentWidth:number;
+    isSuccessWidth:number;
+    isDragEnter:boolean;
+}
+
+const SET_HEIGHT='SET_HEIGHT'
```
ralph-dev, line 38: What is this?

ralph-dev (changes requested):
?

src/Containers/FileUpload/FileUpload.tsx
roggc: this is the redux way to declare types of actions

ralph-dev (approved):
LGTM, will let Andrew approve as well

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -421,15 +432,9 @@ export const FileUpload:React.FC<IFileUploadProps>=({
...
-                <Loading loading={isUploading} message='Uploading...' />
-            </BottomPanel>
-            <BottomPanel withBorder opacity={state.isSuccess.opacity} height={state.loadingContainerHeight} position={state.isSuccess.position} margin={`${state.margin}px`} positionTop={state.positionTopLoading} width={state.isSuccessWidth}>
-                <IsFailureIsSuccessPanel message={successMessage} iconColor={MainTheme.colors.statusColors.green} IconToShow={CheckCircle} />
+            <BottomPanel withBorder padding={isFailure||isSuccess?undefined:'30px 20px 43px 20px'} opacity={state.loading.opacity||state.isSuccess.opacity||state.isFailure.opacity} position={state.loading.position&&state.isFailure.position&&state.isSuccess.position} ref={loadingContainerRef} overflow='hidden' width={isFailure||isSuccess?state.isSuccessWidth:state.componentWidth} margin={`${state.margin}px`} positionTop={state.positionTopLoading} height={isFailure||isSuccess?state.loadingContainerHeight:undefined}>
+                {renderChild()}
```
Ampatte2, line 436: This no longer makes sense to have all these opacities here. As we talked about yesterday this Bottompanel is a container for the children rendered inside of it. Create this panel and transition the opacity and height, it is either displayed or not.

Now, opacity needs to be added to the rendered components. I think this can be achieved by adding a @keyframes to the IsFailureIsSuccessPanel for opacity from 0 to 1 with 0.5s animation. There may be another way to do this.

Padding should also be static, I don't see why it would need to resize.

Add appropriate CSS to BottomPanel/IsFailureIsSuccess so that elements are not at edge of div.

Ampatte2 (changes requested):

Ampatte2:
Also padding and margin from the state should be static variables as they do not change.

ralph-dev (approved):
Still room for improvements but otherwise a LGTM

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -35,6 +33,9 @@ interface IState{
     isDragEnter:boolean;
 }
 
+const padding=10;
+const margin=10;
```
Ampatte2, line 37: PADDING
MARGIN
declare outside component.

Ampatte2 (approved):
Small change and then LGTM

src/Containers/FileUpload/FileUpload.tsx
roggc: just one thing, in the javascript style guide of the docs it says camelCase is permitted when declaring deeply immutable constants.

Ampatte2: https://wolfbeacon.atlassian.net/wiki/spaces/CW/pages/1130102818/JavaScript+Style+Guide#2.4.-Use-Upper-Case-for-Deeply-Immutable-Constants

I would go with the good option rather than permitted if I have a choice?

roggc: in fact in the docs says permitted for uppercase and not required to use uppercase. I think it's a matter of taste

PR #271, opened 2020-12-28
Collapsible-Heading-Molecule

xsharonhe:
I wanted the collapsible header to be extensible, since for the filtering on the left side there are different child elements in the list. Controls also work, but I didn't show that in the gif.
Collapsible-Heading


ralph-dev (approved):
Lgtm

ralph-dev:
I really think for that tag prop we should Maeke it so that
Children OR Data At minimum one must be provided lol but we can quickly look at that later

PR #272, opened 2020-12-29
CEM-1290-Filters-Bar-Organism

xsharonhe:
Actual filtering using react-table will be done with the template.
Filters Bar Organism

### DIFF
--- a/src/Containers/CollapsibleHeading/CollapsibleHeading.tsx
+++ b/src/Containers/CollapsibleHeading/CollapsibleHeading.tsx
@@ -9,7 +9,7 @@ import { Heading } from '@Text/Heading';
 export interface ICollapsibleHeadingProps extends MainInterface,
         ResponsiveInterface,
         React.HTMLAttributes<HTMLDivElement>{
-    title: string;
+    title?: string;
     ChildElement?: React.ReactElement;
     isCollapsed?: boolean;
     setCollapsed?: () => void;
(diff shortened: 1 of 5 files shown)


ralph-dev (approved):
Questions
1) The double indexes, I find questionable
2) Can you justify how this is an organism instead of a molecule just lay out your thoughts 
3) Why is there styling like margin / padding being applied? 
4) Why is the typings not IButton, IHeader, etc

P.S. Feel free to ask me questions if you find anything counter intuitive!

xsharonhe:
> Questions
> 
> 1. The double indexes, I find questionable
> 2. Can you justify how this is an organism instead of a molecule just lay out your thoughts
> 3. Why is there styling like margin / padding being applied?
> 4. Why is the typings not IButton, IHeader, etc
> 
> P.S. Feel free to ask me questions if you find anything counter intuitive!

1. I'm not sure if this second way I did it would be more correct, although not sure why I didn't think of it earlier.
2. Mostly thought of the atoms being the smaller components like (Heading, icon, input, select). The molecule would have been my last ticket (Collapsible Heading), as it involved more than one atom. Since I reused that one molecule multiple times in another component List, I saw it as more of an organism with the same molecule. Hopefully, that explanation makes sense.
3. I forgot to abstract that oops. Think it should be fixed in my new changes.
4. A bit confused by this question. Is this the naming that is the issue? If so, I think this was other people's components and did not want to change how other people named it because it could potentially break other people's applications.

I think I am slightly confused on the need for required props. We discussed this once with the tagProps but also with this one I think I would need to Omit certain props on the FilterSelect because I have already passed in the required props. I didn't do Omit on this one because that would have resulted in a lot of props in Omit and I didn't want to do something that hasn't yet been discussed.

ralph-dev (approved):
Ok explanations look good for this PR let's make sure to discuss it when your done the final template ticket. You can merge this

PR #273, opened 2021-01-01
CEM-1302-Dropdown

arkiis:
Add dropdown and dropdown item components. Add dropdown button as a prop to the dropdown so the user can choose what element triggers the dropdown. 

positoinElement

### DIFF
--- a/src/Containers/Dropdown/Dropdown.tsx
+++ b/src/Containers/Dropdown/Dropdown.tsx
@@ -21,11 +21,15 @@ type DropdownComponent<P = {}> = React.NamedExoticComponent<P> & {
 export interface IDropdownProps extends React.HTMLAttributes<HTMLDivElement> {
     dropdownButton: Element;
     dropdownWidth?: number;
+    right?: boolean;
 };
 
 interface IDropdownContentProps {
     dropdownWidth?: number;
     contentLength: number | undefined;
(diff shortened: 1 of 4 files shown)


src/Containers/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,166 @@
...
+} from 'react';
+import styled from 'styled-components';
+import { transition,scroll } from '../../Utils/Mixins';
+import DropdownItem from './DropdownItem';
+
+type DropdownComponent<P = {}> = React.NamedExoticComponent<P> & {
+  Item: typeof DropdownItem
+}
+
+export interface IDropdownProps {
```
ralph-dev, line 21: Aren't you supposed to be extending the div props here?

ralph-dev (changes requested):
Extend the props

src/Containers/Dropdown/Dropdown.tsx
arkiis: Yeah I just added the props

ralph-dev (approved):
LGTM

ralph-dev (approved):
Placement should affect a fixed type of Left or Right instead of true or false becuase to extend in the future for up and down will require more changes now. It's still good to merge though!

PR #276, opened 2021-01-04
CEM-767-Transaction

Ampatte2:
Not quite sure what the target of this was. Made it more dynamic.

https://wolfbeacon.atlassian.net/jira/software/projects/CEM/boards/6?selectedIssue=CEM-767

I see this video, it looks like it is being controlled by the arrow keys?

How it's working currently is very different. Is it supposed to be a scrollable and onHover do an animation?

CEM-767

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -59,6 +59,7 @@ module.exports = {
         'no-use-before-define': [0],
         'no-underscore-dangle': 'off',
         'no-param-reassign': 'off',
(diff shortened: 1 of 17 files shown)


ralph-dev (approved):
Mixing a PR and Prettier Formatting is hard to read but othewise this is good to merge.
It's supposed to Auto Scroll Up and Down for the Web Layout but the point was totally missed so let's just merge it since its no longer priority anyways :)

PR #281, opened 2021-01-12
Storybook-Babel

Ampatte2:
Npm publish was and still is working, at the meeting tonight I tried to start storybook locally and it failed.

So, turns out storybook also needs it's own babel.

This packs and starts successfully.


ralph-dev (approved):
@arkiis LGTM

PR #283, opened 2021-01-25
CEM-1433 Store Hours List Update

Ampatte2:
Removed a lot of the array finds and replaced with better indexing.

Added formik for handling the forms and errors.

Added moment to handle dates.

Updated Footnote to accept relative positioning for use in other ways than at the bottom of a page.

Now handles no categories and creates a default Regular hours category.
CEM-1433-No Categories

The form handling has been updated for better UX/UI, now with formik the main form will detect changes and allow the user to save on the main page once all changes have been made. They can reset the form to the current initialized value in the state, reinitializing the dirty state.

Formik validation is now used for cleaner error handling.
CEM-1433-Edit Categories

Can reset the form after time has been deleted

CEM-1433-Reset on Delete Time

Now has capability to merge time.
Possible bug or feature. If they select a time period that has the same **to** time and the **new from** time is before the **current from** it will extend times and not conflict. 
Also the same condition but inverted for same **from** time.
CEM-1433-Edit Hours

Merging
CEM-1433-Merging


src/Containers/StoreHoursList/CreateHoursModal.tsx
```diff
@@ -13,223 +15,310 @@ import { Select } from '../../Inputs/Select/Select';
...
         React.HTMLAttributes<HTMLDivElement> {
     isVisible: [boolean, React.Dispatch<React.SetStateAction<boolean>>];
     MODAL_HEADER: string;
     SELECT_A_DAY_TITLE: string;
-    fromTimeTooBigError: string;
-    toTimeTooSmallError: string;
     SELECT_A_CATEGORY: string;
     ADD_HOURS_BUTTON: string;
-    errorMessage: string;
+    FROM_TIME_TO_BIG: string;
```
ralph-dev, line 30: Why not just have on error message?

ralph-dev (commented):
Why not one error message

src/Containers/StoreHoursList/CreateHoursModal.tsx
```diff
@@ -13,223 +15,310 @@ import { Select } from '../../Inputs/Select/Select';
...
+        monday: false,
+        tuesday: false,
+        wednesday: false,
+        thursday: false,
+        friday: false,
+        saturday: false,
+        sunday: false,
+    },
+    storeHours: {
+        from: moment().format(MOMENT_24_HOUR_FORMAT ),
```
ralph-dev, line 60: Remove extra space

src/Containers/StoreHoursList/CreateHoursModal.tsx
```diff
@@ -13,223 +15,310 @@ import { Select } from '../../Inputs/Select/Select';
...
-                            });
-                            setCheckboxes(initialCheckboxState);
-                            return timedArray;
-                        }
-                    }
-                    return null;
-                },
-            );
+        // overWritten Hours are hours that would be lost on merge
+        const overWrittenHoursByDay = {
```
ralph-dev, line 123: Abstract this empty state into an object

src/Containers/StoreHoursList/CreateHoursModal.tsx
```diff
@@ -13,223 +15,310 @@ import { Select } from '../../Inputs/Select/Select';
...
+            wednesday: [],
+            thursday: [],
+            friday: [],
+            saturday: [],
+            sunday: [],
+        }
+
+        // merged Hours are house that will replace overwritten hours
+        // they are the union of overWritten Hours and Store Hours
+        const mergedHoursByDay = {
```
ralph-dev, line 135: Abstract

src/Containers/StoreHoursList/CreateHoursModal.tsx
```diff
@@ -13,223 +15,310 @@ import { Select } from '../../Inputs/Select/Select';
...
+                            storeHoursToMerge.from = hours.from
+                            // check if to (end time) is before current to
+                            if (moment(storeHoursToMerge.to).isBefore(moment(hours.to, MOMENT_24_HOUR_FORMAT ), 'minutes')) {
+                                // sets new end time to current beginning since new falls before
+                                storeHoursToMerge.to = hours.to
+                            }
+                            // copy hours overwritten
+                            overWrittenHoursByDay[day].push(hours);
+                            // Check if new hours enclose current times. The from is before the current from and the to is after the current to
+                        } else  if (moment(storeHoursToMerge.from, MOMENT_24_HOUR_FORMAT ).isBefore(moment(hours.from, MOMENT_24_HOUR_FORMAT ), 'minutes')  && moment(storeHoursToMerge.to, MOMENT_24_HOUR_FORMAT ).isAfter(moment(hours.to, MOMENT_24_HOUR_FORMAT ), 'minutes')) {
```
ralph-dev, line 178: Abstract to a function, lots of reusable code here

ralph-dev (changes requested):
Some code quality feedback, overall very nice merge design

src/Containers/StoreHoursList/CreateHoursModal.tsx
Ampatte2: There is a seperate error message for each I assume to be more customizable, though this was prior to my working on it.

Combining these to an object might make sense but it would be different than the rest of the textHeaders errors so I think it would be okay to leave it this way.

ralph-dev: That doesn't make sense becuase if you have different errors you won't have 1 prop per Type of Error.

src/Containers/StoreHoursList/CreateHoursModal.tsx
Ampatte2: The reason this is like this is it appeared as though all were referencing the same object in memory.

Will use deepCopy to solve this issue.

ralph-dev (approved):
LGTM but logs of warnings / Error Prop duplication is not good

Ampatte2:
The UX was a bit unclear with the save functionality. Now it prompts to save or discard changes when switching categories.
CEM-1433-UX Polish

src/Containers/StoreHoursList/MergeModal.tsx
```diff
@@ -0,0 +1,205 @@
...
+                    storeHours.to,
+                    is24,
+                )}
+            </HoursItem>
+        </Column>
+    )
+    
+    // create list of store hours that require resolution
+    // if no hours require merging return null
+    const renderStoreHoursMerge = () => DAYS_OF_THE_WEEK.map(day => {
```
ralph-dev, line 73: Use JSDOC notation for better readability

ralph-dev (approved):
LGTM, 
The RESET Button UX is a bit confusing? Does that delete everything and reset to an empty state? I wonder if there is a better name for it like, Reset to Empty or Delete All or Reset All or something more clear lol (I don't know the original State it's reseting too basically)

Ampatte2:
Could be named differently. It opens a modal for confirmation that explains more. Could change the modal to say this:
Discard

ralph-dev:
> Could be named differently. It opens a modal for confirmation that explains more. Could change the modal to say this:
> Discard

But what does the reset even do, does it reset current changes as in the changes i made today, or just reset everything back to empty??

PR #284, opened 2021-01-26
CEM-1476 Simple Signup Popup

arkiis:
Pop up that if a user ignores it, the cookie will expire in 30 days, but if the user subscribes, it will expire in one year. 

signupPopup

### DIFF
--- a/src/Containers/SignUpPopup/SignUpPopup.tsx
+++ b/src/Containers/SignUpPopup/SignUpPopup.tsx
@@ -1,0 +1,165 @@
+import React, { ReactElement, useEffect, useState } from 'react'
+import Cookies from 'universal-cookie';
+import { Button } from '@Inputs/Button/Button';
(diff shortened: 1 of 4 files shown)


src/Containers/SignUpPopup/SignUpPopup.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+
+    useEffect(() => {
+        let timer: NodeJS.Timeout;
+        let onCloseTimer: NodeJS.Timeout;
+        if (cookies.get("SignUpPopupSeen")) {
+            return;
+        } 
+        // eslint-disable-next-line prefer-const
+        timer = global.setTimeout(() => {
```
ralph-dev, line 45: Bad name

ralph-dev (approved):
LGTM, great design

PR #286, opened 2021-01-30
Add visit notification

arkiis:
viewnotifcatgion


ralph-dev (approved):
Lgtm, not sure why border radius has important beside it but ok?

PR #288, opened 2021-02-05
CEM-2021 Emoji Mart Build Fix

Ampatte2:
There was a bundling issue with external library css. 

Solution was to use postCss with a rollup plugin for post css. Packed and tested on Dashboard.

Also included text for left header button in vendorFilter.

Error handling was not resetting to fasle on recieving undefined/false for external errors. Now updates appropriately

Before:
EmojiBug

After:
EmojiMart Updated Package


ralph-dev (approved):
LGTM, I did notice that lol I was wondering what happened haha

PR #289, opened 2021-02-08
CEM-2045 box composition

roggc:
do two squares and a rectangle with images that changes coinciding with animation
agorman
as you can see the animation of the text and the squares do not coincide in time.

### DIFF
--- a/src/Containers/BoxComposition/BoxComposition.stories.tsx
+++ b/src/Containers/BoxComposition/BoxComposition.stories.tsx
@@ -9,7 +9,7 @@ export default {
     args: {
         data:[{
             main:{
-                imgSource:['https://dynaimage.cdn.cnn.com/cnn/c_fill,g_auto,w_1200,h_675,ar_16:9/https%3A%2F%2Fcdn.cnn.com%2Fcnnnext%2Fdam%2Fassets%2F210204090136-amanda-gorman-time-cover.jpg'],
+                imgSource:'https://dynaimage.cdn.cnn.com/cnn/c_fill,g_auto,w_1200,h_675,ar_16:9/https%3A%2F%2Fcdn.cnn.com%2Fcnnnext%2Fdam%2Fassets%2F210204090136-amanda-gorman-time-cover.jpg',
                 top:0,
                 left:75,
                 width:100,
@@ -18,7 +18,7 @@ export default {
                 id:'0'
             },
             left:{
-                imgSource:['https://pyxis.nymag.com/v1/imgs/427/8e3/50af665d01b7c7d4339dddc48682d37585-amanda-gorman.rsquare.w700.jpg'],
+                imgSource:'https://pyxis.nymag.com/v1/imgs/427/8e3/50af665d01b7c7d4339dddc48682d37585-amanda-gorman.rsquare.w700.jpg',
                 top:100,
                 left:25,
                 width:75,
@@ -27,7 +27,7 @@ export default {
                 id:'1'
             },
             right:{
-                imgSource:['https://s.abcnews.com/images/Sports/super-bowl-lv-amanda-gorman-01-gty-iwb-210207_1612747303779_hpMain_16x9_992.jpg'],
+                imgSource:'https://s.abcnews.com/images/Sports/super-bowl-lv-amanda-gorman-01-gty-iwb-210207_1612747303779_hpMain_16x9_992.jpg',
                 top:25,
                 left:150,
                 width:75,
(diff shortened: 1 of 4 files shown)


src/Containers/BoxComposition/BoxComposition.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import styled from 'styled-components'
+
+interface IBoxProps{
```
ralph-dev, line 4: Why don't you extend ImageProps???

ralph-dev (changes requested):
You need to pass ...props to all new components / make sure you are extending the correct interface.

src/Containers/BoxComposition/BoxComposition.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Meta, Story } from '@storybook/react';
+import { BoxComposition, IBoxCompositionProps } from '@Containers';
+import { createStoryTitle } from '../../Constants';
+
+export default {
+    title: createStoryTitle('BoxComposition'),
+    component: BoxComposition,
+    args: {
+        data:[
+            {
```
ralph-dev, line 11: This structure won't work, 3 images are part of one set :) try to actually replace the content on the pulse then you will see why :)

src/Containers/BoxComposition/BoxComposition.tsx
```diff
@@ -0,0 +1,81 @@
+import React from 'react'
+import styled from 'styled-components'
+import {SmallText} from '@Text'
+import {Mixins} from '@Utils'
+
+const ANIMATION_DURATION=7
+
+interface IBoxProps extends IImgProps,IDivProps{
+    imgSource:string;
+    withText:boolean;
```
ralph-dev, line 10: Why do you need this?
Just do
!!text &&

ralph-dev (changes requested):
Code structure and props need work, almost there

src/Containers/BoxComposition/BoxComposition.tsx
```diff
@@ -0,0 +1,81 @@
...
+    text:string;
+    id:string;
+}
+
+const Box:React.FC<IBoxProps>=({imgSource,top,left,width,height,withText,text}):React.ReactElement=>
```
Ampatte2, line 15: Why is this a FC and not a function called renderBox = (props) => jsx?

src/Containers/BoxComposition/BoxComposition.tsx
```diff
@@ -0,0 +1,81 @@
...
+    `}
+    border-radius:5px;
+    box-shadow:0 0 1px 0 black;
+`
+
+export interface IBoxCompositionProps{
+    data:IBoxProps[];
+}
+
+export const BoxComposition:React.FC<IBoxCompositionProps>=({data}):React.ReactElement=>
```
Ampatte2, line 76: A FC that just maps? Is there additional styling to be included for this? If not I'd rather just export the Box component and for the story map it out for display.

src/Containers/BoxComposition/BoxComposition.stories.tsx
ralph-dev: [{
mainImage: {imgSources: 'stringUrl',
                top:0,
                left:75,
                width:100,
                height:200,
                text:'bla bla bla',
                id:'0'},
leftImage: {},
rightImage: {}
}, {Set2}]

ralph-dev (changes requested):
Wrong, no need for two arrays.
[
{MainImage: {}, LeftImage: {}, RightImage: {}},
{MainImage: {}, LeftImage: {}, RightImage: {}}
]
Look above ^

src/Containers/BoxComposition/BoxComposition.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { BoxComposition, IBoxCompositionProps } from '@Containers';
+import { createStoryTitle } from '../../Constants';
+
+export default {
+    title: createStoryTitle('BoxComposition'),
+    component: BoxComposition,
+    args: {
+        data:[{
+            main:{
+                imgSource:['https://dynaimage.cdn.cnn.com/cnn/c_fill,g_auto,w_1200,h_675,ar_16:9/https%3A%2F%2Fcdn.cnn.com%2Fcnnnext%2Fdam%2Fassets%2F210204090136-amanda-gorman-time-cover.jpg'],
```
ralph-dev, line 12: imgSource: string
no need for 3 arrays

ralph-dev (changes requested):
All imgSource should be strings no need for the array of strings

ralph-dev (approved):
LGTM, make sure the component still works before merging as the structure was modified.

PR #290, opened 2021-02-13
CEM-2061 box composition 2

roggc:
bc2

### DIFF
--- a/src/Containers/BoxComposition2/BoxComposition2.tsx
+++ b/src/Containers/BoxComposition2/BoxComposition2.tsx
@@ -1,22 +1,17 @@
 import React,{useState,useEffect} from 'react'
 import styled from 'styled-components'
 
-interface IBoxProps{
+interface IBoxProps extends IDivProps,IImgProps{
     imgSrc:string;
-    top:number;
-    left:number;
-    width:number;
-    height:number;
-    borderRadius:number;
 }
 
 export interface IBoxComposition2Props{
-    box1:IBoxProps;
-    box2:IBoxProps;
-    box3:IBoxProps;
+    desktopBox:IBoxProps;
+    mobileBox:IBoxProps;
+    notificationBox:IBoxProps;
 }
 
-export const BoxComposition2:React.FC<IBoxComposition2Props>=({box1,box2,box3}):React.ReactElement=>{
+export const BoxComposition2:React.FC<IBoxComposition2Props>=({desktopBox,mobileBox,notificationBox}):React.ReactElement=>{
     const [showBox2,setShowBox2]=useState(false)
     const [showBox3,setShowBox3]=useState(false)
     useEffect(()=>{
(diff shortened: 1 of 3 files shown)


src/Containers/BoxComposition2/BoxComposition2.tsx
```diff
@@ -0,0 +1,115 @@
...
+    useEffect(()=>{
+        setTimeout(()=>{
+            setShowBox2(true)
+        },1000)
+        setTimeout(()=>{
+            setShowBox3(true)
+        },2000)
+    },[])
+    return     (
+        <div>
```
ralph-dev, line 31: where is the ...props spread here?

src/Containers/BoxComposition2/BoxComposition2.tsx
```diff
@@ -0,0 +1,115 @@
+import React,{useState,useEffect} from 'react'
+import styled from 'styled-components'
+
```
ralph-dev, line 3: You need to extend div for the root prop

src/Containers/BoxComposition2/BoxComposition2.tsx
```diff
@@ -0,0 +1,115 @@
...
+            <Box1 {...box1} />
+            {showBox2&&<Box2 {...box2} />}
+            {showBox3&&<Box3 {...box3} />}
+        </div>
+    )
+}
+
+const Box3:React.FC<IBoxProps>=({imgSrc,top,left,width,height,borderRadius}):React.ReactElement=>
+    (
+        <AnimatedDiv3 top={top} left={left}>
```
ralph-dev, line 41: BAD VARIABLE NAMES 
AnimatedDiv3 is not good!!

src/Containers/BoxComposition2/BoxComposition2.tsx
```diff
@@ -0,0 +1,115 @@
...
+    )
+
+const Box1:React.FC<IBoxProps>=({imgSrc,top,left,width,height,borderRadius}):React.ReactElement=>
+    (
+        <AnimatedDiv1 top={top} left={left}>
+            <Img src={imgSrc} width={width} height={height} borderRadius={borderRadius} />
+        </AnimatedDiv1>
+    )
+
+interface IDivProps{
```
ralph-dev, line 60: Can't this be replaced with HTMLDivProps?

ralph-dev (changes requested):
Code Quality is very bad, name your variables properly. On the bright side the Architecture is correct!

src/Containers/BoxComposition2/BoxComposition2.tsx
roggc: ThirdAnimatedDiv will be ok?

src/Containers/BoxComposition2/BoxComposition2.tsx
roggc: no, I don't think so...

src/Containers/BoxComposition2/BoxComposition2.tsx
roggc: I don't understand...

src/Containers/BoxComposition2/BoxComposition2.tsx
ralph-dev: No. The name has no MEANING.
What is this div?
Is this the CenterDiv
TopLeftDiv etc?
You should be able to read the code and instantly understand what relates to what

src/Containers/BoxComposition2/BoxComposition2.tsx
ralph-dev: You need to do
<div {...props} />
That way when people use this component they can add a margin to it!

ralph-dev (approved):
Looks a lot better just spread the correct props of the root container so another person can add margin when using this componnent

ralph-dev:
You forgot to extend HTMLDIVPROPS to the Prop now that you have ...props btw

PR #292, opened 2021-02-18
CEM-2085 fix visitcount in profilecard position absolute

roggc:
the visitcount in profilecard was positioned absolute which caused to display incorrectly when using in dashboard terminal. adding a position:relative in the parent container should solve the issue (in theory).
Screenshot 2021-02-18 at 11 24 15


ralph-dev (approved):
Lgtm

PR #294, opened 2021-02-19
Cem 2096 default image in profilecard component

roggc:
set a default image on profilecard component
Screenshot 2021-02-19 at 19 59 47


ralph-dev (approved):
Lgtm

PR #295, opened 2021-02-26
CEM-2130 add search functionality to ListHeader component in react-ui

roggc:
lhwithsb

### DIFF
--- a/src/Containers/List/ListHeader.tsx
+++ b/src/Containers/List/ListHeader.tsx
@@ -1,12 +1,12 @@
-import React,{useState} from 'react';
+import React,{useState,useEffect} from 'react';
 import styled from 'styled-components';
 import { StyledIcon } from '@styled-icons/styled-icon';
-import {Search} from '@styled-icons/fa-solid/Search';
-import {Times} from '@styled-icons/fa-solid/Times';
+import {SearchBarExpandable} from '../../Inputs/SearchBarExpandable/SearchBarExpandable';
 import { Heading } from '../../Text/Heading';
-import {Input} from '../../Inputs/Input/Input';
 import { TextLayoutProps } from '../../__Layouts';
 import { Mixins } from '../../Utils';
+
+const DELAY=1000
 
 export interface ListHeaderProps extends TextLayoutProps {
     label?: string;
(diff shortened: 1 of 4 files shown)


Ampatte2:
Video of how this works at 255px width?

src/Containers/List/ListHeader.tsx
```diff
@@ -65,4 +103,15 @@ const Row = styled.div<RowProps>`
...
     ${(props): string | undefined => props.iconProps}
+    cursor:pointer;
 `;
+
+const IconsContainer=styled.div<RowProps>`
+${(props): string | undefined =>
+        props.display && Mixins.flex(props.display)};
+`
+
+const SearchInput=styled(Input)`
```
ralph-dev, line 114: use the Existing React-UI Input

ralph-dev (changes requested):
Expansion Effect is incorrect (It should expand from the Search Icon)
Try Placing a Grey Border around the Icon and Expanding that to get the effect :)

src/Containers/List/ListHeader.tsx
roggc: Yes, it is the existing React-ui input

Ampatte2:
@ralph-dev this style of search is going to be an issue with this component and it's current use cases. It's rarely at fullscreen width. Terminal and dash both use a 255px width for the list.

Possibly make the search drop down below at certain widths?, or transition the search field beneath the header?

ralph-dev:
> @ralph-dev this style of search is going to be an issue with this component and it's current use cases. It's rarely at fullscreen width. Terminal and dash both use a 255px width for the list.
> 
> Possibly make the search drop down below at certain widths?, or transition the search field beneath the header?

This only searches for Order ID, what would be the issue if it takes up the full width of 255 px?

Ampatte2:
It can take up the full 255, right now the search input is inline with the header so it'll look like the small mobile viewport in the second pic.

src/Containers/List/ListHeader.tsx
```diff
@@ -25,31 +29,55 @@ export const ListHeader: React.FC<ListHeaderProps> = ({
...
-            {icon && (
-                <Icon as={icon} onClick={iconClick} iconProps={iconProps} />
-            )}
-            {headerRowComponent}
-        </Row>
-        {children}
-    </Header>
-);
+}): React.ReactElement => {
+    const stateExpanded=useState(false)
```
ralph-dev, line 35: Don't use this , write out both set and the original

ralph-dev (approved):
Lgtm

PR #299, opened 2021-03-03
CEM-2047 Flex Mixin

Ampatte2:
Flex Mixin had reverse-row and reverse-column instead of row-reverse and column-reverse. They have been changed to the correct Mixin.

I think the tablet management PRs did not get prettied so that's why there are so many changed files.


ralph-dev (approved):
LGTM

PR #300, opened 2021-03-03
CEM-2056 IconProps Button

Ampatte2:
Now you can change the size of the icon and set the color of the content.

Maybe color should be renamed => backgroundColor, but I think the contentColor makes that distinction.

Primary and no props will still function the same. Don't know why you'd want to change the color of the primary button? Wouldn't you just use different colors.

CEM-2056-IconProps-Button - No Color or contentColor
CEM-2056-IconProps-Button


ralph-dev (approved):
contentColor is a weird name, buttonContainerColor is more clear or containerColor but ok.

Ampatte2:
Hmmm, it is not clear then. ContentColor is the color of the elements inside the button. If I change color to be more clear to backgroundColor and then contentColor to color I think it would break any existing usages of button that uses color.

ralph-dev:
> Hmmm, it is not clear then. ContentColor is the color of the elements inside the button. If I change color to be more clear to backgroundColor and then contentColor to color I think it would break any existing usages of button that uses color.

Oh contentColor is of the text and icon that makes sense!
Background colour is just for button container background

PR #302, opened 2021-03-04
CEM-2224 Make some changes to SearchBarExpandable

roggc:
searchboxexpandable

### DIFF
--- a/src/Inputs/SearchBarExpandable/SearchBarExpandable.tsx
+++ b/src/Inputs/SearchBarExpandable/SearchBarExpandable.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState,useRef,useCallback } from 'react';
 import { Search } from '@styled-icons/fa-solid/Search';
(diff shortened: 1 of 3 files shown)


src/Inputs/SearchBarExpandable/SearchBarExpandable.tsx
```diff
@@ -30,16 +30,38 @@ export const SearchBarExpandable: React.FC<SearchBarExpandableProps> = ({
         onInput(event.currentTarget.value);
     };
 
+    const inputRef=useRef<HTMLInputElement>(null)
+
+    const expandSearchBox=useCallback(()=>{
+        state[1](true)
```
Ampatte2, line 36: I know modal does this, but modal is not correct do state, setState. It is less props, but it is also confusing to anyone who doesn't understand this.

Ampatte2 (approved):
Great change.

ralph-dev (approved):
LGTM

PR #303, opened 2021-03-04
fixing minwidth in input in searchbar expandable

roggc:
the input field inside SearchBarExpandable had a minimum width which make icon on the right go outside of the box. setting it to 0 solved the problem.
Captura de pantalla 2021-03-04 a las 19 34 02


ralph-dev (approved):
LGTM

PR #304, opened 2021-03-04
adding margin to ListHeader component

roggc:
add margin prop (optional) to ListHeader component


src/Containers/List/ListHeader.tsx
```diff
@@ -52,7 +54,7 @@ export const ListHeader: React.FC<ListHeaderProps>= ({
     },[isExpanded,setShowRest])
 
     return (
-        <Header padding={padding}>
+        <Header padding={padding} margin={margin}>
```
ralph-dev, line 57: Expose a ...props on the root here

ralph-dev (approved):
Could also use ...props to pass margin but this works too

PR #317, opened 2021-03-22
CEM-2329 Fix Search Bar

Ampatte2:
Interesting way to implement this.

Added transitions to ListHeader with search bar instead of removing the element from the dom.

The previous implementation for the search bar was a hack. Now searchbar has a width and an additonal media query.
Expandable searcj


ralph-dev (approved):
LGTM

PR #322, opened 2021-04-01
upgrade babel-plugin-recharts

roggc:
is necessary to upgrade babel-plugin-recharts to last version (2.0.0) in order to storybook load properly (this is because we upgraded to last version of recharts previously)


ralph-dev (approved):
LGTM

PR #329, opened 2021-04-09
Fix-ListHeader-Search

Ampatte2:
https://github.com/cheapreats/react-ui-library/compare/Fix-ListHeader-Search?expand=1#diff-c7f03e5c03d134af5100d8f168484013bd1e659560d132ceecedeaf9817266ebR96

So essentially there was a bouncing effect b/c upon translating back into the component it regains it's relative positioning and moves all elements to the right since the list is not condensed.

This is hacky, but it works. Open to suggestions.

ListHeaderWithSearch

### DIFF
--- a/src/Containers/List/ListHeader.tsx
+++ b/src/Containers/List/ListHeader.tsx
@@ -50,7 +50,7 @@ export const ListHeader: React.FC<ListHeaderProps> = ({
                         {label}
                     </Heading>
                 </HeadingContainer>
-                <SearchContainer isExpanded={isExpanded}>
+                <SearchContainer>
                     {onSearch && (
                         <SearchBarExpandable
                             onInput={onSearch}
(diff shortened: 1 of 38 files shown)


src/Containers/List/ListHeader.tsx
```diff
@@ -82,21 +82,32 @@ interface IResponsiveSearchProps {
...
     ${Mixins.transition(['transform', 'opacity'])}
+    @keyframes heading {
+        0% {
+            position: fixed;
+        }
+        90% {
+            position: fixed;
+        }
+        100% {
+            position: relative;
```
ralph-dev, line 104: Why is there position fixed in the first place?

ralph-dev (approved):
Lgtm, not sure why there is position fixed in first place. If you can explain that we can solve it

src/Containers/List/ListHeader.tsx
Ampatte2: To remove it from the stacking context?

## Without Fixed
Out of stacking

PR #330, opened 2021-04-09
On stop draggable

rogers-ch:
image

We added props and functions so that when you click on a table on the EditDraggableCanvas, the arrayIndex is passed up.

We also added a function to pass the new coordinates up after a table is dragged to a new location on the EditDraggableCanvas.

If you look at the console, you can see that when a table is clicked, the coordinates and arrayIndex are printed to the console. This is just for testing to show that our function wiring is working.


ralph-dev (approved):
Lgtm

PR #333, opened 2021-04-12
Absolute Positioning Fix

dboichuk:
Made the positions for tables absolute. This allows the tables to be added on top of each other (stack up) without pushing them down.


ralph-dev (approved):
Lgtm

PR #334, opened 2021-04-13
Masked Input - Fix Final

Ampatte2:
Much much better UI/UX and useability:

Trying to figure out how to point people to the docs for react input mask for how to configure the mask, could do presets as well?

No longer need length

## $$$
Input Mask number

## Phone
Phone


ralph-dev:
@Ampatte2 Can you take a video of toggling between Percentage and Amount

Ampatte2:
## Multiple Masks
Multiple Masks

ralph-dev (approved):
LGTM

ralph-dev:
Remove log from code and show a video before I can approve btw

Ampatte2:
It just doesn't work, html events are always converted to strings, and getting this done it's going to just expose the value. Would be great if it could use the event but it cannot.

PR #336, opened 2021-04-17
fix file upload

roggc:
https://drive.google.com/file/d/1Y37N0EZau_G8RZ27VDjF06MLKDDcgVLD/view?usp=sharing

now setIsUploading, setIsFailure and setIsSuccess functions are passed as props to the component to control the status of uploading (reading) a file content

### DIFF
--- a/src/Containers/FileUpload/FileUpload.stories.tsx
+++ b/src/Containers/FileUpload/FileUpload.stories.tsx
@@ -32,3 +32,31 @@ export const Basic: Story<IFileUploadProps> = (args) => {
         <FileUpload {...args} {...componentProps} />
     )
 };
+
+export const VeryLongMessageDuration: Story<IFileUploadProps> = (args) => {
+    const [isUploading,setIsUploading]=useState(false)
+    const [isSuccess,setIsSuccess]=useState(false)
+    const [isFailure,setIsFailure]=useState(false)
+
+    const componentProps={
+        isUploading,isSuccess,isFailure,setIsUploading,setIsSuccess,setIsFailure,messageDuration:20000
+    }
+
+    return (
+        <FileUpload {...args} {...componentProps} />
+    )
+};
+
+export const VeryShortMessageDuration: Story<IFileUploadProps> = (args) => {
+    const [isUploading,setIsUploading]=useState(false)
+    const [isSuccess,setIsSuccess]=useState(false)
+    const [isFailure,setIsFailure]=useState(false)
+
+    const componentProps={
+        isUploading,isSuccess,isFailure,setIsUploading,setIsSuccess,setIsFailure,messageDuration:200
+    }
+
+    return (
+        <FileUpload {...args} {...componentProps} />
+    )
+};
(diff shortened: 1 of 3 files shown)


src/Containers/FileUpload/FileUpload.stories.tsx
```diff
@@ -13,15 +13,50 @@ export default {
...
+    const componentProps={
+        isUploading,isSuccess,isFailure,setIsUploading,setIsSuccess,setIsFailure
+    }
+
+    return (
+        <FileUpload {...args} {...componentProps} />
+    )
+};
+
+export const VeryLongMessageDuration: Story<IFileUploadProps> = (args) => {
```
ralph-dev, line 36: Why duplicate entire code? Why not just pass duration as a prop?

ralph-dev (changes requested):
MessageDuration should be a prop

src/Containers/FileUpload/FileUpload.stories.tsx
roggc: I have done it like this because passing it as a prop in args doesn't work with storybook (there is no difference when changing the value in storybook controls) . so putting it this way, with different stories, it's how you see the difference that this value make on the behaviour of the component

roggc: messageDuration it's a prop, it's an optional prop

ralph-dev: It does work in storybook, you need to define subComponentProps to access in storybook

ralph-dev (changes requested):
Still not ready, let's do a call tomorrow if your still having trouble. In the meantime, try fixing the other Excel issues

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -466,8 +482,20 @@ export const FileUpload: React.FC<IFileUploadProps> = ({
                         );
                     }
                     if (base64StringFile) {
-                        setBase64(base64StringFile);
+                        setIsSuccess(true)
+                        setIsFailure(false)
+                        setIsUploading(false)
+                        setTimeout(()=>{
+                            if(isMounted.current)
```
ralph-dev, line 489: Put brackets around if statements!

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -466,8 +482,20 @@ export const FileUpload: React.FC<IFileUploadProps> = ({
...
+                                setIsSuccess(false)
+                            setBase64(base64StringFile)
+                        },messageDuration)
```
ralph-dev, line 492: is this only SuccessMessageDuration or includes errors as well?

ralph-dev (approved):
Lgtm

src/Containers/FileUpload/FileUpload.tsx
roggc: includes both isSuccess and isFailure message.

PR #337, opened 2021-04-18
Repushing the fix for the bug for the Horizontal tables

ScottyKnows44:
finished the user story for adding the horizontal rectangle table to the canvas for the CheaprEats Terminal. In order to complete user story I needed to remove the set width in the _DraggableTable.tsx file. This allows all tables to be created with their proper width.


ralph-dev (approved):
Lgtm

PR #338, opened 2021-04-18
fix-fileupload-2

roggc:
now shows name of file being uploaded-read.
fixfileupload2


ralph-dev (approved):
LGTM, I would've included
File Name | File Size | A Cancel Upload Button (Just incase file is crashing browser)

PR #339, opened 2021-04-19
fileupload-cancel-btn

roggc:
add cancel button to fileupload component

https://drive.google.com/file/d/1vmNjkmsGTXz7vEB0lmtjdLo6Qfu_nHe4/view?usp=sharing


src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -488,39 +489,28 @@ export const FileUpload: React.FC<IFileUploadProps> = ({
...
-                            if(isMounted.current){
-                                setIsSuccess(false)
-                            }
-                        },messageDuration)
-                    }
+            const worker=new Worker('./worker.js', { type: 'module' })
+            workerRef.current=worker
+            worker.postMessage({file})
+            worker.onmessage=(e)=>{
+                const {base64StringFile}=e.data
```
ralph-dev, line 496: {data}
Then
Const {base64} = data

Works as well but your way is fine

src/Containers/FileUpload/FileUpload.tsx
```diff
@@ -488,39 +489,28 @@ export const FileUpload: React.FC<IFileUploadProps> = ({
...
+                if(base64StringFile!==undefined){
```
ralph-dev, line 497: What if it equals null?
Use 
!Var

ralph-dev (approved):
Lgtm, nice job this is a very important concept to pick up when doing intensive tasks on the web!

src/Containers/FileUpload/FileUpload.tsx
roggc: but what if is an empty string because file read was empty?

roggc: maybe and can use ??

roggc: I don't think it can be null... can be undefined or empty string (or normal string)

PR #340, opened 2021-04-19
Added new prop toolbarUse that will change box-shadow css

dboichuk:
The new prop will determine if the table is used in the addTableToolbar. If it is used in the toolbar it will not highlight red when the table is clicked.

### DIFF
--- a/src/Containers/CircleTable/CircleTable.tsx
+++ b/src/Containers/CircleTable/CircleTable.tsx
@@ -60,7 +60,7 @@ export interface ICircleTable {
     /**
      * Determines if the table is used in the toolbar or not
      */
-    toolbarUse?: boolean;
+    isInAddTableToolbar?: boolean;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


src/Containers/CircleTable/CircleTable.tsx
```diff
@@ -57,6 +57,10 @@ export interface ICircleTable {
      * @param selectedChildIndex - the array index for the table
      */
     onTableClick: (selectedChildIndex: number) => void;
+    /**
+     * Determines if the table is used in the toolbar or not
+     */
+    toolbarUse?: boolean;
```
ralph-dev, line 63: isHighlightedWhenFocused

Is the better prop name to use. Always is prefix for Boolean

ralph-dev (approved):
Lgtm naming for Boolean should always prefix with IS

src/Containers/CircleTable/CircleTable.tsx
dboichuk: Sound good! I will fix it and make another PR

PR #341, opened 2021-04-21
changing worker library

roggc:
with the actual implementation, the fileupload component does not work when using in dashboard project. changing the worker library for workerize-loader that does not need to add plugins to webpack maybe helps in fileupload component working for dashboard project


ralph-dev (approved):
Lgtm

PR #346, opened 2021-04-23
using workerize-loader

roggc:
despite the name of the branch, I ended up using workerize-loader for the webpack experience (storybook uses webpack and dashboard too) and rollup-plugin-copy to copy the worker file into the dist folder, because in this way dashboard project will find it. I have tested and it works in both projects. of course, people should do npm install after fetching this branch.


ralph-dev (approved):
You've tested this?

roggc:
yes, in my computer doesn't give problems. people should do npm install. if it's the header maps thing what give problems to people when doing npm install, then they should do 'npm config set @here:registry https://repo.platform.here.com/artifactory/api/npm/maps-api-for-javascript/' first

ralph-dev:
Okay let's merge this then and wrap up the task today

ralph-dev:
Book time for a call to review the dashboard once it's ready

ralph-dev (approved):
LLGTM

PR #347, opened 2021-04-23
Added onClickfunctions for the editControlPanel

ScottyKnows44:
I added the onClick functions for both the - and + buttons for the EditControlPanel. This pairs up with the event listeners I have created on the Edit page for the terminal. The purpose being to change the amount of chairs a table has when the user selects them.


src/Containers/EditControlPanel/EditControlPanel.stories.tsx
```diff
@@ -23,6 +24,14 @@ const handleCloneClick = () => {
...
 
+const handleRemoveChairClick = () => {
+    console.log('- Button has been clicked');
+};
+
+const handleAddChairClick = () => {
+    console.log('+ Button has been clicked');
+};
+
 const Template: Story<IEditControlPanel> = (args) => (
```
ralph-dev, line 35: why not just use Args Types so it shows up in the STorybook Actions (Makes it eaiser to debug)

ralph-dev (approved):
LGTM but use ArgTypes optionally if you wnat.

https://storybook.js.org/docs/react/api/argtypes

Look at Button Componet how it uses it btw

src/Containers/EditControlPanel/EditControlPanel.stories.tsx
ScottyKnows44: Good idea, I will convert to use Arg Types after I finish the user story to edit the amount of chairs for a table

PR #349, opened 2021-04-25
Adding more table types

ScottyKnows44:
Added squareTableTypes with the purpose of stating what the is purpose of each square table. That way when adding and deleting chairs they will always remain the same shape.


ralph-dev (approved):
Lgtm

PR #351, opened 2021-04-26
Eye icon substitution

rogers-ch:
Changed Chair icons on the edit page from plus/minus to Eye/EyeSlash:

image


ralph-dev (approved):
Why are all the eyes different?

PR #352, opened 2021-04-30
Chair refactoring

dboichuk:
In order to update the chair state in the main table array, I had to go add three new props to a chair and make some refactoring.


ralph-dev (approved):
Lgtm

PR #353, opened 2021-04-30
Added back on focus feature that got lost

dboichuk:


ralph-dev (approved):
Lgtm

ralph-dev:
I would've used a more generic name like isHighlightedWhenSelected

Since what if you need it off in multiple places

PR #354, opened 2021-05-05
Table focus work

rogers-ch:
Added selectedIndex prop and useEffect so the currently-selected table will maintain focus on re-render


src/Containers/CircleTable/CircleTable.tsx
```diff
@@ -82,10 +86,27 @@ export const CircleTable: React.FC<ICircleTable> = ({
...
+     * Use useEffect to keep focus on TableBody after re-render if the
+     * selectedIndex number matches the arrayIndex number for this table
+     */
+    useEffect(() => {
+        if (selectedIndex === arrayIndex) {
+            if (tableBodyRef != null) {
+                tableBodyRef.current.focus();
+            }
+        }
+    });
```
ralph-dev, line 108: No dependency array will re render every second

ralph-dev (approved):
Lgtm but watch that useEffect will hit preformance

rogers-ch:
Requesting re-review before merging into master. I made a few changes last week after you reviewed this PR while I was trying to fix the bug we told you about last Friday. All of those changes cleared all checks last week and published to npm. 

Today I just fixed the conflict in the package.json file. 

There are errors in the library coming from the Invite and CustomSearch components. Those are not affiliated with our project, and it looks like the errors were introduced sometime in the last day.

ralph-dev (approved):
Lgtm

PR #357, opened 2021-05-11
added location prop to ReservationSideBar component and bumped up the version number...

ScottyKnows44:
I have created two components on the Management page in figma called PartyInfoInput and SeatingInfoInput. This involved me adding a location prop to the IReservationList and IWaitingListList. I have included pictures as well.

image
image


package.json
```diff
@@ -119,6 +119,7 @@
         "babel-plugin-styled-components": "^1.12.0",
         "babel-plugin-transform-class-properties": "^6.24.1",
         "emoji-mart": "^3.0.0",
+        "fa-icons": "^0.2.0",
```
ralph-dev, line 122: Why?

package.json
```diff
@@ -129,6 +130,7 @@
         "react-beautiful-dnd": "^13.0.0",
         "react-dropzone": "^11.2.4",
         "react-easy-flip": "^4.0.0-beta.2",
+        "react-icons": "^4.2.0",
```
ralph-dev, line 133: Use styled icons?

ralph-dev (changes requested):
Don't add more libraries to the repo if the existing ones can solve your problem

PR #358, opened 2021-05-12
Input field work

dboichuk:
I finished working on refactoring EditControlPanel. I added a new prop function that will be triggered when the user changes the Table Number. This new prop will be used in EditPage in the terminal to change the Table Number for tables when the user changes it on the control panel. 

I have used --no-verify to commit because I added a bunch of files with errors that I have not touched to my commit for some reason. I have no idea how to fix those files to I used no verify.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@cheapreats/react-ui",
-    "version": "2.4.15",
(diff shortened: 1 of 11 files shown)


ralph-dev (approved):
Lgtm

PR #362, opened 2021-05-17
bumped version to 2.4.24

cmarro3:
bumped the version to 2.4.24 so then it could publish to the cheapreats react UI npm


ralph-dev (approved):
Lgtm

PR #368, opened 2021-10-08
(CEM-2504) Highlighted text component

frootloopers:
Finished the functionality of the HighlightedText component and added a story for it; Added an export to ListItem.tsx for VerticalListItemProps (renamed from ListItemProps to avoid collision). 

highlight1 highlight2

### DIFF
--- a/src/Containers/HighlightedText/HighlightedText.stories.tsx
+++ b/src/Containers/HighlightedText/HighlightedText.stories.tsx
@@ -1,10 +1,10 @@
 import React from 'react';
 import { Meta, Story } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
-import { HighlightedText, HighlightedTextProps } from '../../index';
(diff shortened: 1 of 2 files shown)


ralph-dev:
> Hey Ralph, I tried using the component from DropDown.tsx, and I'm having a couple of issues.
> 
> 
> 
> Firstly, I can't figure out why there's an error when I try to create a Dropdown with just a dropdownButton as an attribute when it worked in the Dropdown.stories.tsx.
> 
> image
> 
> 
> 
> Secondly, how can I make the dropdownButton appear inline with the text?
> 
> 

You don't have the right closing tag here that's why

frootloopers:
> You don't have the right closing tag here that's why

Here is more of the code:
image
The tags should be correct. Dropdown is imported from src/Containers/Dropdown/index. Not sure what IntrinsicAttributes is and why VSCode is making a big deal out of it. And I'm confused as to why doing this fixes that error:
image

frootloopers:
New appearance using Dropdown instead of the normal List:
image

PR #371, opened 2021-10-12
CEM-2509:Add a Prop to Table Component

Jwunder2:
"Added timeLastServed variable with formatting of Hours:Minutes:Seconds in both CircleTable and SquareTable"


ralph-dev:
Attach a photo to the Pr

Jwunder2:
I changed the code to only display timer when the condition of occupied is met.
TableOccupied
TableVacant

ralph-dev:
fix conflicts

PR #372, opened 2021-10-15
CEM-2521 Analytics Component

Madmanmayson:
Created an Analytics Component which can automatically scale the number down to a reasonably viewable size using the common numeric suffixes 

image


ralph-dev:
Ready to merge

PR #380, opened 2021-10-22
Cem 2516

percypie:


src/Containers/LoyaltyPoints/LoyaltyPoints.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react'; 
+import styled from 'styled-components';
+import { Stars } from '@styled-icons/material/Stars';
+import {Paragraph} from '@Text';
+
+export interface LoyaltyPointsProps
+    extends React.HTMLAttributes<HTMLDivElement>{
+            LoyaltyAmount: number;
```
ralph-dev, line 8: Why are props uppercase? With no docs

ralph-dev (changes requested):
Please include image in PR

src/Containers/LoyaltyPoints/LoyaltyPoints.tsx
percypie: Im having an issue with the text colour being forced to be the colour of paragraph rather than my styled div since changing to it, cant figure out how to fix that.

percypie:
Screenshot 2021-11-07 104825

ralph-dev:
Have you tried overriding the styles? If so, maybe you need to specify a higher specificity

percypie:
Yeah I tried this but it didnt work, is there another way to specify a higher specificity. also I've talked to Hunter about how he did it using styled.header, but from my understanding that doesnt work does it? Im pretty sure its Heading and not header, and you have to override it with brackets not a '.' right?
Screenshot 2021-11-07 180557

ralph-dev:
@percypie 
In this case let's not re-use the paragraph component then, I'll explain why on our team call.

percypie:
alright thank you, sorry I couldnt figure it out.

percypie:
image

PR #385, opened 2021-10-29
Highlighted Text Part 3

frootloopers:
image

When new text is added to labels, the previous text automatically becomes transparent. The 'age' of text can also be overridden to force them to appear as 'old' or 'new'.

### DIFF
--- a/src/Containers/HighlightedText/HighlightedText.tsx
+++ b/src/Containers/HighlightedText/HighlightedText.tsx
@@ -8,6 +8,11 @@ import { TextLayoutProps } from '../../Fragments/TextLayout';
 const newTextOpacity = 1;
 const oldTextOpacity = 0.5;
 
+enum Age{
+    New = 0,
+    Old = 1,
+}
+
 export interface HighlightedString {
     /** contents of the string */
     text: string;
@@ -21,8 +26,8 @@ export interface HighlightedString {
     listItemsArgs?: Array<IDropdownItemProps>;
     /** content of DropdownItems */
     listItemsBodies?: Array<JSX.Element>;
-    /** true to override age as 'new', false to override as 'old'. Leave unfilled for automatic. */
-    overrideAge?: boolean;
+    /** New (0) for opaque text, Old (1) for transparent text. None for automatic */
+    overrideAge?: Age;
 }
 
 // extends line div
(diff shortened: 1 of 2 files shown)


src/Containers/HighlightedText/HighlightedText.tsx
```diff
@@ -19,6 +21,8 @@ export interface HighlightedString {
     listItemsArgs?: Array<IDropdownItemProps>;
     /** content of DropdownItems */
     listItemsBodies?: Array<JSX.Element>;
+    /** true to override age as 'new', false to override as 'old'. Leave unfilled for automatic. */
+    overrideAge?: boolean;
```
ralph-dev, line 25: Why not use am Enum instead of this indirect mapping

ralph-dev (changes requested):

src/Containers/HighlightedText/HighlightedText.tsx
```diff
@@ -1,10 +1,17 @@
...
 import styled from 'styled-components';
 import { ClickableSmallText, SmallText } from '@Text';
 import Dropdown, { IDropdownProps } from '../Dropdown/Dropdown';
 import DropdownItem, { IDropdownItemProps } from '../Dropdown/DropdownItem';
 import { TextLayoutProps } from '../../Fragments/TextLayout';
 
+const newTextOpacity = 1;
+const oldTextOpacity = 0.5;
+
+enum Age{
```
ralph-dev, line 11: Why are you using an Enum here, isn't the number anything in between 0-1

ralph-dev (changes requested):

src/Containers/HighlightedText/HighlightedText.tsx
ralph-dev: This comment was meant for when you didn't have manual override fyi

src/Containers/HighlightedText/HighlightedText.tsx
```diff
@@ -36,42 +45,62 @@ export const HighlightedText: React.FC<HighlightedTextProps> = ({
     labels,
     ...props
 }): React.ReactElement => {
+    const [numEntries, setNumEntries] = useState(0);
+    const [numNewEntries, setNumNewEntries] = useState(0);
+
+    if (labels.length !== numEntries) {
+        setNumNewEntries (labels.length - numEntries)
+        setNumEntries(labels.length)
+    }
```
ralph-dev, line 54: Why is this not in a function? It will execute every re render . Maybe put it in a useEffect?

ralph-dev (changes requested):

ralph-dev (approved):
Lgtm

PR #387, opened 2021-10-29
Cem 2567

mmendoza1964:
CEM-2567

Added TableHeaderCell component, to be implemented into other bigger table components.
Contains a title for the cell and a arrow that displays if a filter is passed in.

tableHeaderCell-preview1


ralph-dev (approved):
Lgtm, arrow should have an up and down state you can edit the CSS to rotate the arrow 180 btw

PR #390, opened 2021-11-03
CEM-2569 CRMTable Component

Madmanmayson:
So I've gotten the filtering implemented into the CRMTable component, though it requires either declaring whether each column is or isn't filtered and how so or by passing in a default column filter as an optional prop. However I've been having some issues dealing with the TypeScript errors and typing the properties for filtering without it just being completely generic such as any for defaultColumn. On top of that, the mock filter component I created in the stories file for creating a filter box has properties throwing additional typescript errors (There are also a bunch that aren't actually errors as I was using a bunch of components created by teammates that will be merged in from master). But beyond the typescript errors I believe the table is complete.

NOTE: due to the object style for data in the Component based story filtering doesn't actually work as I used react-table's built in filter that only work on primitive types rather then writing my own in the story file.

Raw Text Example
image

With Components Example:
image


src/Containers/CRMTable/CRMTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+    useFilters,
+    useTable
+} from 'react-table';
+import { CRMRowProps } from '../CRMRow/CRMRow';
+
+
+export interface ICRMTableProps extends React.HTMLAttributes<HTMLTableElement> {
+    data: Array<CRMRowProps>; 
+    columns: Array<Column<CRMRowProps>>;
+    defaultColumn?: any,
```
ralph-dev, line 14: Don't use the any type

src/Containers/CRMTable/CRMTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+        rows, 
+        prepareRow 
+    } = useTable({ columns, data, defaultColumn }, useFilters);
+
+    const buildHeaderGroups = () => headerGroups.map((headerGroup) => (
+        <CRMTableHeaderRow {...headerGroup.getHeaderGroupProps()}>
+            {headerGroup.headers.map((column) => (
+                <CRMTableHeaderCell {...column.getHeaderProps()}>
+                    {column.render('Header')}
+                    <div>{column.canFilter ? column.render('Filter') : null}</div>
```
ralph-dev, line 38: Use && versus ? : null

ralph-dev (commented):
It's on its way

src/Containers/CRMTable/CRMTable.tsx
Madmanmayson: The issue I've had with the defaultColumn is that its an object with two fields at least and I can't seem to find any prebuilt type in react-table that can be used for it like I had for Column. So should I create my own class to represent the fields inside of the object or what. Because for filtering the default options you need to give for every column is "Filter" which is the component to render for the filter and "filter" which is the function used for filtering each with their own types.

Though I might also be able to move away from having default columns in general but in that case, every column has to declare if its filtered or not.

src/Containers/CRMTable/CRMTable.tsx
Madmanmayson: this part should be fixed now

src/Containers/CRMTable/CRMTable.tsx
Madmanmayson: Alright it should be properly typed now

src/Containers/CRMTable/CRMTable.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+    CRMTable, 
+    CustomerProfile,
+    ICRMTableProps, 
+    TableHeaderCell,
+    TagGroup
+} from '../../index';
+import { createStoryTitle } from '../../Constants';
+
+interface FilterProps {
+    column: any;
```
ralph-dev, line 14: not updated, let me know when y ou want me to re-revie wbtw

ralph-dev (changes requested):
Let me know when you. want me to re-review

Madmanmayson:
Storybook types should be fixed and should be ready for a re-review now

src/Containers/CRMTable/CRMTable.stories.tsx
Madmanmayson: resolved using a built in react-table interface.

ralph-dev (approved):
LGTM, run 'npx tsc' and solve any typescript errors if you have any

Madmanmayson:
Typescript errors fixed

PR #391, opened 2021-11-03
Added Header Component along with subcomponents (OrderStatus, Payment…

Emil-Ilnicki:
…Method, and OrderType). Also refactored code from OrderDetailsComponent branch namely rendertext -> cardtext


Emil-Ilnicki:
Header
Header component with passed in JSX components to be rendered top left, bottom left, top right, and bottom right

Emil-Ilnicki:
PaymentMethod
Payment method component restricted and formatted with enums

Emil-Ilnicki:
OrderStatus
Order status component restricted with OrderStatus enums and color enums

Emil-Ilnicki:
OrderType
Order type component restricted with order type enums

Emil-Ilnicki:
OrderItem
Refactored RendText.jsx component now called CardText.tsx used in OrderDetailsContent.jsx

src/Containers/Header/Header.tsx
```diff
@@ -0,0 +1,54 @@
...
+import styled from 'styled-components';
+
+export interface IHeaderProps {
+    /* string or JSX component for top left of header */
+    topLeft?: string | JSX.Element;
+    /* string or JSX component for bottom left of header */
+    bottomLeft?: string | JSX.Element;
+    /* string or JSX component for top right of header */
+    topRight?: string | JSX.Element;
+    /* string or JSX component for bottom right of header */
```
ralph-dev, line 11: Don't repeat the type in the description

src/Containers/Header/OrderDetailsHeader.stories.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+import OrderStatus from '@Containers/OrderStatus/OrderStatus';
+import OrderType from '@Text/OrderType/OrderType';
+import OrderPaymentMethod from '@Text/OrderPaymentMethod/OrderPaymentMethod';
+import { Paragraph } from '@Text/Paragraph';
+import Header from './Header';
+import { createStoryTitle } from '../../Constants';
+
+const topLeftElement = (
```
ralph-dev, line 10: Why is this variable here?

src/Containers/OrderStatus/OrderStatus.tsx
```diff
@@ -0,0 +1,46 @@
...
+    PLACED = 'PLACED',
+    PREPARING = 'PREPARING',
+    PREPARED = 'PREPARED',
+    COMPLETE = 'COMPLETE',
+    CANCELLED = 'CANCELLED',
+}
+
+interface IOrderStatusProps {
+    /* Status of order (Placed, Preparing, Cancelled etc...) */
+    orderStatus: string;
```
ralph-dev, line 14: This is type Enum not a string

ralph-dev (changes requested):
Changes

Emil-Ilnicki:
orderpaymentcard
orderpaymentcard2
Order payment card with different story instances showcasing that this component is dynamic

Emil-Ilnicki:
orderTemplate
ordertemplatephone
Order template with mock data slotted into its components. Also addressed non-responsive order payment card with media queries to go to width: 100% on smaller view ports

src/Containers/Header/Header.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Paragraph } from '@Text/Paragraph';
+import Header from './Header';
+import { createStoryTitle } from '../../Constants';
+
+export default {
+    title: createStoryTitle('Order Header'),
+    component: Header,
+    args: {
+        topLeft: (
+            <Paragraph size="1.4rem" bold>
```
ralph-dev, line 21: Suggestion: if you were to make this an OrderID component you can use the slice() logic inside the component instead of repeating it in your application. I would do this for maintainability sake

src/Containers/OrderStatus/OrderStatus.tsx
```diff
@@ -0,0 +1,46 @@
...
+    /* Status of order (Placed, Preparing, Cancelled etc...) */
+    orderStatus: OrderStatusIdentifier;
+}
+
+export const OrderStatus = ({
+    orderStatus,
+    ...props
+}: IOrderStatusProps): React.ReactElement => {
+    /* Returns color based on order status */
+    const handleOrderStatus = () => {
```
ralph-dev, line 22: Missing return type

src/Text/OrderPaymentMethod/OrderPaymentMethod.tsx
```diff
@@ -0,0 +1,46 @@
...
+    /* Type of payment (Credit card, wallet, in person) */
+    paymentMethod: OrderPaymentMethodTypes;
+}
+
+export const OrderPaymentMethod: React.FC<IPaymentMethodProps> = ({
+    paymentMethod,
+    ...props
+}) => {
+    /* Returns payment method with no underscores */
+    const handlePaymentMethod = () => {
```
ralph-dev, line 26: Return type

src/Text/OrderType/OrderType.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const OrderType: React.FC<IOrderTypeProps> = ({
+    orderType,
+    ...props
+}): React.ReactElement => {
+    /* Returns order type with no underscores */
+    const handleOrderType = () => {
+        switch (orderType) {
+            case OrderTypeIdentifier.EAT_IN:
+                return orderTypes.EAT_IN;
```
ralph-dev, line 37: Why not just return the enum!?

ralph-dev (approved):
Lgtm but fix the return enum thing

src/Text/OrderType/OrderType.tsx
Emil-Ilnicki: How would I remove the underscore then? Should I just reuse the regex from the terminal repo and create a separate helper function for it?

src/Containers/Header/Header.stories.tsx
Emil-Ilnicki: Would having a styled component with these values be a suitable alternative? Where then people can just pass in style overrides through ...props, cause I'm not sure how slice() logic would look like in this case.

ralph-dev: Styled Component couldn't contain the slice logic defeating the whole point of letting OrderID Component own the logic.

ralph-dev: Image across Terminal and your ordering service repeating the .slice() logic versus just importing an OrderID component that does it once instead

src/Text/OrderType/OrderType.tsx
ralph-dev: Yep I would use a helper function. Becuase your basically doing the same thing with that object but it's excessive if you think about how many times you might have to do this. 
1 Function versus hundreds of objects.

PR #393, opened 2021-11-05
CEM-2582 Responsive Tags

B-Fuqua:
Modified the tag group component to be responsive, as well as implementing a fade effect so the user knows to scroll through the tags when applicable.
image


ralph-dev (approved):
Lgtm

PR #400, opened 2021-11-11
GR-CRM CEM-2587 Row Selector

B-Fuqua:
Modified the CRM Table component adding checkboxes to select specific rows.
image


B-Fuqua:
There are linter warnings on line 37, and 48 but changing those lines broke the code entirely. I'm not sure how I need to work around them.

ralph-dev (approved):
Lgtm

PR #412, opened 2021-11-19
CEM-2596 Draggable Row

3p1cPh4r40h:
image
Initial state

image
While dragging

image
After being switched

A draggable row based on the MiddleCanvas component logic for reordering react-draggable items.


ralph-dev (approved):
LGTM

PR #416, opened 2021-11-20
Attempting to fix the lost files in master

Madmanmayson:
As of right now CircleTable.tsx is the one reporting errors and it looks completely different from the one found in the current state of master despite having no changes in the past 24 days on the main master branch so I'm not sure if the broken version is the correct version and such.


ralph-dev (approved):
Lgtm

PR #423, opened 2021-11-24
Lakehead - CEM-2602 Profile list with counter

percypie:
image

### DIFF
--- a/src/Containers/FeaturedProfilesCount/FeaturedProfilesCount.tsx
+++ b/src/Containers/FeaturedProfilesCount/FeaturedProfilesCount.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import styled from 'styled-components';
+import  {IFeaturedProfilesCardProps}  from 'index';
+
+export const FeaturedProfilesCount: React.FC<IFeaturedProfilesCardProps> = ({
+    profileData, //the array of profiles to be displayed.
+    ...props
+}): React.ReactElement => (
+    <CountContainer>{profileData.length}</CountContainer>
+)
+
+const CountContainer = styled.span`
+    position: relative;
+    top: -105px;
+    left: 660px; 
+    color: black;
+    font-size: 50px
+    
+`;
+
+//todo Make location change with amount of profiles (currently having issues trying to make Featured profile card not generate a new line without being able to edit it)
--- a/src/Containers/FeaturedProfilesCount/FeaturedProfilesCount.stories.tsx
+++ b/src/Containers/FeaturedProfilesCount/FeaturedProfilesCount.stories.tsx
@@ -1,0 +1,98 @@
+import React from 'react'; 
+import { Meta, Story } from '@storybook/react';
+import { createStoryTitle } from '../../Constants';
+import { FeaturedProfilesCard, IFeaturedProfilesCardProps, FeaturedProfilesCount} from "index";
+
+const profiles = [
+    {
+        image:
+            'https://images.pexels.com/photos/220453/pexels-photo-220453.jpeg?cs=srgb&dl=pexels-pixabay-220453.jpg&fm=jpg',
+        initials: 'AC',
+        id: 2345,
+    },
+    {
+        image:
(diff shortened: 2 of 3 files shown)


src/Containers/FeaturedProfilesCount/FeaturedProfilesCount.tsx
```diff
@@ -0,0 +1,21 @@
...
+const CountContainer = styled.span`
+    position: relative;
+    top: -105px;
+    left: 660px; 
+    color: black;
+    font-size: 50px
+    
+`;
+
+//todo Make location change with amount of profiles (currently having issues trying to make Featured profile card not generate a new line without being able to edit it)
```
ralph-dev, line 21: ?

ralph-dev (changes requested):
Not approving since todo let

ralph-dev (commented):
Why not update the original component to have this feature?

percypie:
I assumed I wasn't allowed to edit the existing component, if I can edit the existing component then I can fix this very easily

src/Containers/FeaturesProfileCard/FeaturedProfilesCard.tsx
```diff
@@ -90,14 +90,19 @@ export const FeaturedProfilesCard: React.FC<IFeaturedProfilesCardProps> = ({
...
-        <Container>
-            {renderProfileCircles()}
-            <FeaturedProfile icon key={ADD_USER_ICON_KEY} background="grey" />
-        </Container>
+        <Holder>
+            <Container>
+                {renderProfileCircles()}
+                <FeaturedProfile icon key={ADD_USER_ICON_KEY} background="grey" />
+            </Container>
+            <CountContainer>{profileData.length}</CountContainer>
```
ralph-dev, line 98: This needs to be optional

Prop && Jsx

ralph-dev (changes requested):
What if the user doesn't want this. When you edit the existing component you need to make sure not to break existing users workflows ;)

src/Containers/FeaturesProfileCard/FeaturedProfilesCard.tsx
```diff
@@ -20,10 +20,13 @@ export interface IProfile {
 
 export interface IFeaturedProfilesCardProps {
     profileData: IProfile[];
+    counterBool: boolean;
```
ralph-dev, line 23: Documentation of Props goes here not on the component. You are doucmenting the type

src/Containers/FeaturesProfileCard/FeaturedProfilesCard.tsx
```diff
@@ -20,10 +20,13 @@ export interface IProfile {
...
 }
 
 export const FeaturedProfilesCard: React.FC<IFeaturedProfilesCardProps> = ({
-    profileData,
+    profileData, //the array of profiles to be displayed
+    counterBool, //if true, the profile counter is displayed, and if false, the counter is not displayed
```
ralph-dev, line 28: docs should not be on this line, put in TypeScript

ralph-dev (approved):
Fix the Comments before merging, LGTM

src/Containers/FeaturesProfileCard/FeaturedProfilesCard.stories.tsx
```diff
@@ -87,15 +87,26 @@ const profiles = [
...
+const Template: Story<IFeaturedProfilesCardProps> = (args) => (
     <FeaturedProfilesCard {...args} />
 );
 
-//todo different use cases presets.
\ No newline at end of file
+export const Basic = Template.bind({});
+Basic.args = {
+    profileData: profiles,
+    counterBool: false,
```
ralph-dev, line 99: Don't put Types in the Name

We dont say NameString, we just say Name.

Rename this to isCounterShowing or isCounterDisplayed or isProfileCountShowing

the prefix is denotes a Boolean value

ralph-dev (approved):
LGTM, I would update variable name

PR #426, opened 2021-12-01
Green River David - CEM-2600 - Reservation Calendar Upgrade

pavldavid:
Upgrade the existing calendar in the React UI Library to optionally support Prices under the Days of the Month.
Prices could be color (Green, Yellow or Red) 

Green - Good Price
Yellow - Okay Price
Red - Surge Price

CEM2600


ralph-dev (changes requested):
New fields should be optional by default. Also how come it's showing the same price for everyday?

ralph-dev:
Attach a screenshot with different prices under each day / different Colors

PR #430, opened 2022-01-08
Tag improvement

ericlin11354:
Implementing iconBehaviour

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cheapreats/react-ui",
-  "version": "2.4.52",
+  "version": "2.4.53",
   "description": "React UI library for CheaprEats",
   "main": "./dist/index.js",
   "module": "./dist/index.esm.js",
--- a/src/Containers/Tag/Tag.tsx
+++ b/src/Containers/Tag/Tag.tsx
@@ -18,14 +18,13 @@ export interface TagProps extends ITagDiv {
 export const Tag: React.FC<TagProps> = ({
     icon = Times,
     children,
-    isHoverable = true, 
-    iconVisible = true,
+    isHoverable = true,
     iconBehaviour = 'Hover',
     ...props
 }): React.ReactElement => (
     <TagDiv {...props} isHoverable={isHoverable}>
         {children}
-        {iconVisible && <Icon as={icon} iconBehaviour={iconBehaviour} />}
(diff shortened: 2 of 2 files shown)


ericlin11354:
Here is a demo of the Tag component. 
https://user-images.githubusercontent.com/50276107/148658238-556c2319-fe03-4dab-b71b-3853a14b20d6.mp4
Notice at the end of the video the tag does not change when transition from state 'Hover' to 'None'. This might be due to the fact that the hover effect has been permanently written to all TagDiv children. Ideally, this should not be a problem as we're likely to not switch between states with existing Tags (If that's the case, a simple fix could be to recreate the Tag with the desired state).

ericlin11354:
@ralph-dev 
I took your suggestion and displayed certain css within the hover effect based on iconBehaviour state. However, the issue remains unfixed. I believe it has to do with Storybook's rendering. Demonstrated [here](https://user-images.githubusercontent.com/50276107/148662086-b838c1a0-1f83-4864-85ac-5ff6b1dfd6f5.mp4), None type works when the page is refreshed with iconBehaviour === 'None'. Otherwise, it fails to disable the 'X'. 

Alternatively, I attempted to resolve the issue via rewriting the css used in the hover effect. Demonstrated [here](https://user-images.githubusercontent.com/50276107/148662501-a87208eb-c538-4a82-9076-9b5baad077d4.mp4), this results in hover effect of the first transition staying (i.e. Hover -> None means None hover stays and vice-versa).

Code Snippet for both attempts:
Attempt 1:
```
${({ iconBehaviour }): string =>
    iconBehaviour === 'Always' ? `
        width: 10px;
        margin-left: 10px;
    ` : `
        ${TagDiv}:hover & {
            ${iconBehaviour === 'Hover' ? `
                width: 10px;
                margin-left: 10px;
            ` : ``
            }
        }
    `
}
```
Attempt 2:
```
${({ iconBehaviour }): string =>
    iconBehaviour === 'Always' ? `
        width: 10px;
        margin-left: 10px;
    ` : `
        ${TagDiv}:hover & {
            ${iconBehaviour === 'Hover' ? `
                width: 10px;
                margin-left: 10px;
            ` : iconBehaviour === 'None' ? `
                width: 0;
                margin-left: 0;
            ` : ``
            }
        }
    `
}
```

PR #435, opened 2022-01-17
roger- no ticket - Drop area

roggc:
create dropArea component that uses dropzone component

dropAreav6
disabled state:
Screenshot 2022-01-18 at 15 12 56
small mobile:
Screenshot 2022-01-19 at 12 50 19


src/Containers/DropArea/DropArea.stories.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+import { DropArea, IDropAreaProps } from './DropArea';
+
+export default {
+    title: 'Components/DropArea',
+    component: DropArea,
+    args: {},
+    argTypes: { onClick: { action: 'clickeddd!!!' } },
```
ralph-dev, line 9: There are actions missing here

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import styled from 'styled-components';
+import { MainInterface, Main } from '@Utils/BaseStyles';
+
+export interface IDropAreaProps
+    extends MainInterface,
+        React.HTMLAttributes<HTMLDivElement> {
+    message?: string;
+    isDragEnter?: boolean;
```
ralph-dev, line 9: onDragEnter
onDragLeave
As functions instead

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,62 @@
...
+    onClick?: () => void;
+}
+
+export const DropArea: React.FC<IDropAreaProps> = ({
+    message = 'Drag and drop your files or click here to select',
+    isDragEnter = false,
+    padding = '10px',
```
ralph-dev, line 16: CSS should not be here put this default value in the styled component

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,62 @@
...
+            }
+          }
+          background-image: linear-gradient(90deg, #3399ff 50%, transparent 50%), linear-gradient(90deg, #3399ff 50%, transparent 50%), linear-gradient(0deg, #3399ff 50%, transparent 50%), linear-gradient(0deg, #3399ff 50%, transparent 50%);
+          background-repeat: repeat-x, repeat-x, repeat-y, repeat-y;
+          background-size: 15px 2px, 15px 2px, 2px 15px, 2px 15px;
+          background-position: left top, right bottom, left bottom, right   top;
+          animation: border-dance .3s infinite linear;
+`
+        : `
+border:2px dashed grey;
```
ralph-dev, line 56: Colours use the theme or add a new colour to the theme

ralph-dev (changes requested):
Fix CSS and events

src/Containers/DropArea/DropArea.tsx
roggc: but I am not using the dropzone component in this component. do you want me to use this library for this component (dropzone)?

ralph-dev: Yes you can use the drag part of the libarary

roggc: I mean, to handle the files and all this...

roggc: ok I will try and see what happens...

ralph-dev: No do not move the logic for handling files. It's just a DragEvent like I listed in Teams. 

Don't put workers here! The file can be dropped then just an action is emitted of file details that's it

src/Containers/DropArea/DropArea.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+    args: {},
+} as Meta;
+
+export const Basic: Story<IDropAreaProps> = (args) => {
+    const [isDragEnter, setIsDragEnter] = useState(false);
+    const onDragEnter = () => {
+        setIsDragEnter(true);
+    };
+    const onDragLeave = () => {
+        setIsDragEnter(false);
```
ralph-dev, line 17: Remove this prop it's not needed

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,95 @@
...
+        | (<T extends File>(
+              acceptedFiles: T[],
+              fileRejections: FileRejection[],
+              event: DropEvent,
+          ) => void)
+        | undefined;
+    isDragEnter?: boolean;
+}
+
+export const DropArea: React.FC<IDropAreaProps> = ({
```
ralph-dev, line 26: Should be in styled component

ralph-dev (changes requested):
Still missing action emitters. Don't console log the file, try using the action :)

src/Containers/DropArea/DropArea.tsx
roggc: no, I need to pass isDragEnter as a prop

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,95 @@
...
+    message = 'Drag and drop your files or click here to select',
+    onDragEnter=()=>null,
+    onDragLeave=()=>null,
+    onDropHandler=()=>null,
```
ralph-dev, line 30: Remove word handler it's implied

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,95 @@
...
+    padding,
+    ...props
+}): React.ReactElement => {
+    const theme = useTheme();
+    const defaultPadding = useMemo((): string | number => {
+        if (padding === undefined) {
+            return theme.dimensions.padding.container;
+        }
+        return padding;
+    }, [padding, theme]);
```
ralph-dev, line 41: Move to styled component

src/Containers/DropArea/DropArea.stories.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+import { DropArea, IDropAreaProps } from './DropArea';
+
+export default {
+    title: 'Components/DropArea',
+    component: DropArea,
+    args: {},
+    argTypes: {
```
ralph-dev, line 9: Where is on Button Click, look at the picture I sent you, let's improve the UI

ralph-dev (changes requested):

src/Containers/DropArea/DropArea.tsx
roggc: no, becuase it collisions with onDrop attribute of HTML div.

ralph-dev: Then if it's already there why are you overriding it?

If you still need to override, do onFileDrop

src/Containers/DropArea/DropArea.stories.tsx
roggc: there is no onClick, it's handled automatically by the input element ( <input {...getInputProps()}/>) where getInputProps comes from dropzone.

roggc: and also by the div which contains the input (<div {...getRootProps()} />), where getRootProps comes again from dropzone

src/Containers/DropArea/DropArea.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const Basic: Story<IDropAreaProps> = (args) => <DropArea {...args} />;
+export const IsDragEnterTrue = Basic.bind({});
+IsDragEnterTrue.args = {
+    ...Basic.args,
+    isDragEnter: true,
+};
+export const OnClickPreventDefault = Basic.bind({});
+OnClickPreventDefault.args = {
+    ...Basic.args,
+    onClick: (e: React.MouseEvent) => {
```
ralph-dev, line 26: What is this doing here? Do you mean disabled state?

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,140 @@
...
+    ...props
+}): React.ReactElement => {
+    const { getInputProps, getRootProps } = useDropzone({
+        onDragEnter,
+        onDragLeave,
+        onDrop: onDropHandler,
+    });
+    return (
+        <Dropzone multiple>
+            {() => (
```
ralph-dev, line 44: Why is there a function here?

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,140 @@
...
+                <div
+                    {...getRootProps({
+                        onClick: (e: React.MouseEvent) => {
+                            e.stopPropagation();
+                        },
```
ralph-dev, line 49: Do you mean CSS pointer events none??

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,140 @@
...
+                    })}
+                >
+                    <DropAreaBox isDragEnter={isDragEnter} {...props}>
+                        <MessageBox>{message}</MessageBox>
+                        <OrBox>OR</OrBox>
+                        <BrowseFiles>
+                            browse files
```
ralph-dev, line 56: Should be a prop

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,140 @@
...
+                            <input
+                                {...getInputProps({
+                                    onClick,
+                                    style: {
```
ralph-dev, line 60: Styled component

ralph-dev (commented):
Almost there

ralph-dev:
Let's add the icon / make the UI prettier like the animation etc

src/Containers/DropArea/DropArea.stories.tsx
roggc: this is to test if we can prevent default behaviour, and we can

src/Containers/DropArea/DropArea.tsx
roggc: because this is render props pattern and is how it works with the Dropzone component.

src/Containers/DropArea/DropArea.tsx
roggc: it doesn't work with styled component, I tryied

src/Containers/DropArea/DropArea.tsx
roggc: I don't know what you mean. this is to restrict click behaviour to the button.

src/Containers/DropArea/DropArea.stories.tsx
ralph-dev: You should test disabled state not this. This is not a state.

src/Containers/DropArea/DropArea.tsx
ralph-dev: Can't it be abstracted outside the tree?

src/Containers/DropArea/DropArea.tsx
ralph-dev: Google Pointer Events CSS "Pass Through :)

src/Containers/DropArea/DropArea.tsx
ralph-dev: It does work, don't include the style part

ralph-dev (changes requested):
Needs work

src/Containers/DropArea/DropArea.stories.tsx
roggc: if we have an onClick prop, we must test it

roggc: I have also tested the isDisabled state

src/Containers/DropArea/DropArea.tsx
roggc: it doesn't

src/Containers/DropArea/DropArea.tsx
roggc: I don't think so, you would be abstracting all the tree

src/Containers/DropArea/DropArea.stories.tsx
ralph-dev: > if we have an onClick prop, we must test it

You test the onClick prop using action. 

And you test whether onClick is not emitting that action when in DISABLED state

src/Containers/DropArea/DropArea.tsx
ralph-dev: Ok let's keep it then I guess

src/Containers/DropArea/DropArea.tsx
ralph-dev: It does. I will call you later tonight once you have fixed the other issues and show you how. Paste the error here in the meantime

src/Containers/DropArea/DropArea.stories.tsx
roggc: I have tested that, I wanted to test if I was able to prevent default behaviour. it's a test, it's not harmfull

src/Containers/DropArea/DropArea.tsx
roggc: what error?

ralph-dev:
Fix the height of the icon so when the animation plays the things isn't popping up and down

src/Containers/DropArea/DropArea.stories.tsx
ralph-dev: This is not a good test. Testing disabled state is better becuase it will test this behaviour as well. Remove this test, it's bloat.

Also style the disabled state to gray out the UI a bit

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,179 @@
...
+import { flex } from '@Utils/Mixins';
+import Dropzone, {
+    useDropzone,
+    DropEvent,
+    FileRejection,
+} from 'react-dropzone';
+import Lottie from 'react-lottie';
+import { animationData } from './animationData';
+
+const lottieOptions = {
```
ralph-dev, line 14: A recurring file animation for when the user is dragging 

Docs

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,179 @@
...
+const RootDiv = styled.div`
+    width: fit-content;
+`;
+
+const Input = styled.input`
+    opacity: 0;
+    display: initial !important;
+    visibility: initial;
+    position: absolute;
+    z-index: 99999999;
```
ralph-dev, line 102: This is all not needed. Put pointer Event to pass through all parents or whatever is capturing the event and then you can remove this hack

ralph-dev (changes requested):
Almost done

src/Containers/DropArea/DropArea.tsx
roggc: but then the dragging behaviour doesn't work

ralph-dev: It will work, dragging behaviour is not affected by this

roggc: yes, it is! it stops working! I tell you. If I have done this change (go back to input styled component) it's because dragging behaviour was not working, and the reason for this was the CSS property pointer-events:none; I can tell because when I removed that property it started to work again the dragging behaviour. I think it's because the on-hover event, that is not detected when pointer-events:none;

roggc: I will try some things but I suppose I will not succeed...

roggc: I tried and did not succeed. The only way to achieve the desired result is as it is now (with the hacky CSS).

ralph-dev: Use a different property other than none. Like pass through . None will not work

roggc: there not exist this property for pointer-events

roggc: the thing is when the click event reaches RootDiv, it triggers onclick event on input element. we don't want this behavior. it can be canceled by assigning an onClick handler to the RootDiv, as I have done assigning stopPropagation function.

ralph-dev (commented):
Need to Fix Popping of Height due to animation.

src/Containers/DropArea/DropArea.tsx
roggc: maybe I can avoid the hacky css by using a reference to the input element and triggering the click event in the onclick handler of the browseFiles element

roggc: I can do without the hacky css but paying a price, the price is you cannot preventDefault behaviour, so this means you cannot prevent from the dialog box to select a file to appear (that's why it's important to test and hence the importance of the story you told me to delete). do you want me to keep the hacky css and being able to prevent default behaviour or you want to loose this ability and have the code without the hacky css?

ralph-dev: You can prevent at the function layer

ralph-dev: Just do it the correct way I will show you how, you call some code before you call the omClick event

roggc: https://stackoverflow.com/q/66621550/12441000

roggc: when you dispatch an event programmatically you cannot prevent default behavior

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,197 @@
...
+            return <Lottie options={lottieOptions} width={140} height={ICON_HEIGHT+DIFFERENCE_BETWEEN_ICON_HEIGHT_AND_LOTTIE_HEIGHT} />;
+        return <Icon as={CloudUploadAlt} />;
+    };
+
+    const stopPropagation = (e: React.MouseEvent): void => {
+        e.stopPropagation();
+    };
+
+    return (
+        <Dropzone multiple>
```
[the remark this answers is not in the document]
roggc: no I cannot put props here because it doesn't extend htmldivelement props

roggc:
> Need to Fix Popping of Height due to animation.

the popping effect of height occurs even without the icon or lottie animation, it is due to the border animation

ralph-dev:
It's popping becuase of the Animation size you can fix it easily

ralph-dev:
> It's popping becuase of the Animation size you can fix it easily

If it's due to the animation of the border why would that modify the height don't modify the height

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,215 @@
...
+                new MouseEvent('click', { bubbles: false }),
+            );
+        }
+    };
+
+    useEffect(() => {
+        if (inputRef.current) {
+            inputRef.current.onclick = onClickHandler;
+        }
+    }, [onClickHandler]);
```
ralph-dev, line 93: What's this for?

src/Containers/DropArea/DropArea.tsx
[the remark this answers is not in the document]
ralph-dev: Then extend the correct thing...

src/Containers/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,215 @@
...
+        if (isDragEnter)
+            return (
+                <Lottie
+                    options={lottieOptions}
+                    width={140}
+                    height={ICON_HEIGHT}
+                />
+            );
+        return <Icon as={CloudUploadAlt} />;
+    };
```
ralph-dev, line 79: UseCallback or Memo here I think your Gif is very laggy all of a sudden?

src/Containers/DropArea/DropArea.tsx
roggc: this is the only way to assign onclick handler to input element

src/Containers/DropArea/DropArea.tsx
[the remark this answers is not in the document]
roggc: it's a component from the library

src/Containers/DropArea/DropArea.tsx
ralph-dev: html input already has native onClick... remove this

src/Containers/DropArea/DropArea.tsx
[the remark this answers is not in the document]
ralph-dev: Then grab the type from the libarary. Not everything is a div element

src/Containers/DropArea/DropArea.tsx
roggc: it doesn't work. that's the reason I have done it like this. I already had it like you say and had to change because didn't work

src/Containers/DropArea/DropArea.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+BigWidth.args = {
+    ...Basic.args,
+    width: 500,
+};
+export const IsDisabledTrue = Basic.bind({});
+IsDisabledTrue.args = {
+    ...Basic.args,
+    isDisabled: true,
+};
+export const TestPreventDefault = Basic.bind({});
```
ralph-dev, line 28: Remove this... the disabled state tests this

ralph-dev (commented):
Can you screenshot what this component looks like on Mobile view

src/Containers/DropArea/DropArea.stories.tsx
roggc: no. this is not true.

ralph-dev: What is the user test you are trying to do here?

roggc: when you click the Browse Files button, the default behaviour is to open dialog box to choose file. with the hacky CSS I can, and anyone using this component can, prevent this default behavior to happen (that is, not open dialog box). without the hacky CSS I can't, and always will open this dialog box, apart from doing whatever other thing you program on the onClick event handler (everything except preventing this default behavior). the disabled state removes any interaction with the component. so essentially suppose a developer wants to do something else than opening the dialog box when the user clicks, that is, show a message instead or whatever other thing but not open the dialog box to choose file. with the actual implementation he can't, but with the hacky CSS he can. disabled state doesn't allow this because remove any kind of interaction (all interactions).

roggc: so essentially the test or story I am trying to implement is TestDoNotOpenTheDialgoBoxToChooseFileButDoThisOtherThingInsteadWhenClicking

ralph-dev (approved):
Lgtm, let's move on to the list of panels

PR #439, opened 2022-01-22
roger- no ticket - create panelListWrapper component

roggc:
rec2
rec1

### DIFF
--- a/src/Containers/PanelListWrapper/PanelListWrapper.tsx
+++ b/src/Containers/PanelListWrapper/PanelListWrapper.tsx
@@ -45,7 +45,7 @@ export const PanelListWrapper: React.FC<IPanelListWrapperProps> = ({
     return (
         <PanelListWrapperBox {...props}>
             {fadeOutPanels.map(({ panel, isShown }) => (
-                <PanelCardFadeOut
+                <PanelCardFadeInFadeOut
                     key={`${panel.name}`}
                     {...panel}
                     margin={`${verticalSpacing}px 0`}
@@ -75,7 +75,7 @@ const PanelCardFadeIn = styled(PanelCard)<{ fadeIn: number }>`
     animation: fadein ${({ fadeIn }) => fadeIn}ms;
 `;
 
-const PanelCardFadeOut = styled(PanelCardFadeIn)<{
+const PanelCardFadeInFadeOut = styled(PanelCardFadeIn)<{
     isShown: boolean;
     fadeOut: number;
 }>`
--- a/src/Containers/PanelListWrapper/PanelListWrapper.tsx
+++ b/src/Containers/PanelListWrapper/PanelListWrapper.tsx
@@ -23,14 +23,14 @@ export const PanelListWrapper: React.FC<IPanelListWrapperProps> = ({
     delaySequentially = 100,
     ...props
 }) => {
-    const internalPanels = useSequentiallyAddedPanels(
+    const sequentiallyPanels = useSequentiallyAddedPanels(
         panels,
         delaySequentially,
     );
     const panelsToMap = useMemo(
         (): IPanelCardProps[] =>
-            isAddPanelsSequentially ? internalPanels : panels,
-        [isAddPanelsSequentially, internalPanels, panels],
+            isAddPanelsSequentially ? sequentiallyPanels : panels,
+        [isAddPanelsSequentially, sequentiallyPanels, panels],
     );
 
     return (
--- a/src/Containers/PanelListWrapper/PanelListWrapper.tsx
+++ b/src/Containers/PanelListWrapper/PanelListWrapper.tsx
@@ -3,8 +3,6 @@ import React, { useEffect, useState, useMemo, useRef } from 'react';
 import styled from 'styled-components';
 import { PanelCard, IPanelCardProps } from '../PanelCard/PanelCard';
 
-const DELAY_IN_MOUNTING_IN_MS = 100;
-
 export interface IPanelListWrapperProps
     extends MainInterface,
         React.HTMLAttributes<HTMLDivElement> {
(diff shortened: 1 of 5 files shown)


src/Containers/PanelListWrapper/PanelListWrapper.tsx
```diff
@@ -0,0 +1,18 @@
+import { MainInterface,Main } from '@Utils/BaseStyles';
+import React from 'react';
+import styled from 'styled-components';
+import {PanelCard,IPanelCardProps} from '../PanelCard/PanelCard';
+
+export interface IPanelListWrapperProps extends MainInterface,React.HTMLAttributes<HTMLDivElement>{
+    panels?:IPanelCardProps[];
```
ralph-dev, line 7: Panels is required. It can be an empty array but it should never be null right? Unless you are supporting the no data available but I don't see that in your code

src/Containers/PanelListWrapper/PanelListWrapper.tsx
```diff
@@ -0,0 +1,18 @@
...
+    /** vertical space between panels in px */
+    verticalSpacing?:number;
+}
+
+export const PanelListWrapper:React.FC<IPanelListWrapperProps>=({panels,verticalSpacing,...props})=><PanelListWrapperBox {...props}>
+    {panels?.map((panel)=><PanelCard key={`${panel.name}`} {...panel} margin={`${verticalSpacing}px 0`} />)}
+</PanelListWrapperBox>
+
+const PanelListWrapperBox=styled.div<MainInterface>`
+${(props)=>Main({...props})}
```
ralph-dev, line 17: Why?

ralph-dev (changes requested):

src/Containers/PanelListWrapper/PanelListWrapper.tsx
roggc: this is for the padding and margin props passed to the component as props instead of inside style prop

src/Containers/PanelListWrapper/PanelListWrapper.tsx
roggc: well , the code handles the undefined case with a panels?.map() .... doesn't crash (Basic story)

ralph-dev: But it doesn't show any visual to the user. This is the same state as an empty array then...

src/Containers/PanelListWrapper/PanelListWrapper.tsx
```diff
@@ -0,0 +1,80 @@
...
+        [isAnimatedPanels, internalPanels, panels],
+    );
+
+    return (
+        <PanelListWrapperBox {...props}>
+            {panelsToMap.map((panel) => (
+                <PanelCard
+                    key={`${panel.name}`}
+                    {...panel}
+                    margin={`${verticalSpacing}px 0`}
```
ralph-dev, line 36: Move it to the Styled Components

src/Containers/PanelListWrapper/PanelListWrapper.tsx
```diff
@@ -0,0 +1,80 @@
...
+};
+
+const PanelListWrapperBox = styled.div<MainInterface>`
+    ${(props) => Main({ ...props })}
+`;
+
+const useAnimatedPanels = (panels:IPanelCardProps[]) => {
+    const [internalPanels, setInternalPanels] = useState<IPanelCardProps[]>([]);
+    const [panelIndex, setPanelIndex] = useState<number>(0);
+    const panelsLength = useMemo(() => panels.length, [panels]);
```
ralph-dev, line 50: move inside function

src/Containers/PanelListWrapper/PanelListWrapper.tsx
```diff
@@ -0,0 +1,80 @@
...
+            ))}
+        </PanelListWrapperBox>
+    );
+};
+
+const PanelListWrapperBox = styled.div<MainInterface>`
+    ${(props) => Main({ ...props })}
+`;
+
+const useAnimatedPanels = (panels:IPanelCardProps[]) => {
```
ralph-dev, line 47: JSDOC here explaining in English the Logic Behaviour you are applying
Load Panels one by one with a Fade In Animation.
Load Panels one by one with a Fade Out Animation

src/Containers/PanelListWrapper/PanelListWrapper.tsx
roggc: it's more work, I have to do a styled component with an interface, call this styled component passing as prop verticalSpacing prop, so here it's the same but instead of doing all of the above I only pass margin prop. that's why we define maininterface and use it in components

src/Containers/PanelListWrapper/PanelListWrapper.tsx
roggc: ?? I deleted the use of panelsLength as you said, and use panels.length instead

src/Containers/PanelListWrapper/PanelListWrapper.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export interface IPanelListWrapperProps
+    extends MainInterface,
+        React.HTMLAttributes<HTMLDivElement> {
+    panels: IPanelCardProps[];
+    /** vertical space between panels in px */
+    verticalSpacing?: number;
+    /** if true panels are added sequentially to the list */
+    isAddPanelsSequentially?: boolean;
+    /** dealy in ms for each panel added to the list */
```
ralph-dev, line 15: delay

src/Containers/PanelListWrapper/PanelListWrapper.tsx
```diff
@@ -0,0 +1,63 @@
...
+    delaySequentially?: number;
+    /** fade in in ms */
+    fadeIn?:number;
```
ralph-dev, line 18: What about fadeOut time?

ralph-dev (changes requested):
Still missing

ralph-dev:
Also the box grows by popping, try to grow the box height instead

src/Containers/PanelListWrapper/useFadeOut.ts
```diff
@@ -0,0 +1,63 @@
...
+        isShown: boolean;
+    }[],
+    () => void,
+] => {
+    const previousPanelsAmmount = useRef<number>();
+    const [internalPanels, setInternalPanels] = useState(
+        panels.map((panel) => ({ panel, isShown: true })),
+    );
+
+    useEffect(() => {
```
ralph-dev, line 23: JSDOC

ralph-dev (changes requested):
Need to fix a few things and add an animation to the container list itself to grow using an animation instead of popping the height

ralph-dev (approved):
Lgtm

PR #443, opened 2022-01-28
CEM-2599

iamjada:
CEM-2599 
Removed unused table props and added initials on top of chair.

Screen Shot 2022-01-27 at 5 24 08 PM


ralph-dev (changes requested):
Why don't only the top chair leg have props and not the bottom leg? Better yet, why not just have ChairLegProps so they stay in sync ;)

iamjada:
Added ChairLegProps to top and bottom chair legs. Created a Personalized Chair with square legs.
Screen Shot 2022-01-27 at 7 48 53 PM

ralph-dev (approved):
Lgtm, I would just add back chair props and seat props as well

ralph-dev:
backChairProps

Als ways add the s

ralph-dev:
Also why do you need a div and chair body can one of them be removed?

PR #444, opened 2022-01-28
GRC-CEM-2608-SearchableDropdown

Jwunder2:
SearchableDropdown component- I got the dropdown and the searchbar working. It isn't as flexible as its intended to be but I wanted to get a review for the main part of the component.

image
image


ralph-dev:
isMounted not mount

PR #446, opened 2022-01-28
Yurii - CEM-2716 -  Table Selection Style

Zhuk314:
Capture
Capture2


src/Containers/SquareTable/SquareTable.stories.tsx
```diff
@@ -472,6 +472,6 @@ BarRectangleTable.args = {
     ],
     isSquare: false,
     tableUse: 'TableForManagement',
-    onTableClick: handleTableClick,
+    onTableClick: action("Clicked table with index: "),
```
ralph-dev, line 475: Use argTypes so you don't need to add the import btw

ralph-dev (approved):
Lgtm

ralph-dev:
You may want to copy the File Upload V2 components Border Animation btw that way there is some micro interactions here for clarity

PR #447, opened 2022-01-28
roger- Cem 1199- restaurant review card

roggc:
https://wolfbeacon.atlassian.net/browse/CEM-1199
Screenshot 2022-01-28 at 09 51 13

### DIFF
--- a/src/Containers/RatingItem/RatingItem.tsx
+++ b/src/Containers/RatingItem/RatingItem.tsx
@@ -64,6 +64,7 @@ const Label = styled.div`
     font-weight: 700;
     font-size:${({theme})=>theme.font.size.small};
     opacity:0.7;
+    width:30px;
 `;
 
 const RatingBar = styled(RB)`
--- a/src/Containers/RatingItem/RatingItem.tsx
+++ b/src/Containers/RatingItem/RatingItem.tsx
@@ -55,7 +55,7 @@ export const RatingItem: React.FC<IRatingItemProps> = ({
                 thickness={barThickness}
                 {...getBarColorProp()}
             />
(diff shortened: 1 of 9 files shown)


src/Containers/RatingItem/RatingItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+        default:
+            return '?';
+        }
+    }, [stars]);
+
+    const innerRating = useMemo(() => {
+        if (rating > 1) return 1;
+        if (rating < 0) return 0;
+        return rating;
+    }, [rating]);
```
ralph-dev, line 50: Function is repeated across components?

src/Containers/RatingItem/RatingItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    const getBarColorProp = useCallback((): object => {
+        if (barColor) return { color: barColor };
+        return {};
+    }, [barColor]);
```
ralph-dev, line 55: CSS

ralph-dev (approved):
Lgtm but code quality needs imrpovement

src/Containers/RatingItem/RatingItem.tsx
roggc: yes, it is in two components, in RatingBar and RatingItem. RatingItem makes use of RatingBar, but it also has a rating label besides of the ratingbar. That's why is repeated, because the label's value also needs to be adjusted.

src/Containers/RatingItem/RatingItem.tsx
roggc: color it is a prop

